PR #4, opened 2022-08-18
Create codeql-analysis.yml

gruyaume:
Create codeql-analysis.yml

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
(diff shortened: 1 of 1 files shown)


ghislainbourgeois (approved):
LGTM

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '42 11 * * 3'
```
sanchezfdezjavier, line 21: I guess this date is just random for a weekly check. Is that right?

gruyaume: Yes, the cron job will run at 11:42 on Wednesdays. This is an automatically generated file btw, we could change it to whatever else we want.

PR #17, opened 2023-01-20
CI fixes

sanchezfdezjavier:
# Description

Solves several CI issues by refactoring the integration tests.

# Checklist:


tests/integration/test_integration.py
```diff
@@ -93,7 +75,16 @@ async def deploy_charm(ops_test, charm: Path) -> None:
...
+    @pytest.mark.abort_on_fail
+    @pytest.fixture(scope="module")
+    async def build_and_deploy(self, ops_test):
+        ops_test.destructive_mode = False
+        charm = await ops_test.build_charm(".")
+        await self.deploy_charm(ops_test, charm)
+
+    @pytest.mark.abort_on_fail
+    @pytest.fixture(scope="module")
+    async def post_deployment_tasks(self, ops_test) -> str:
```
Gmerold, line 87: Add typing to `ops_test` and update docstring to reflect argument change.

tests/integration/test_integration.py
```diff
@@ -114,15 +105,17 @@ async def post_deployment_tasks(self, namespace: str) -> str:
         generated_token = vault.generate_token(ttl="5m")
         return generated_token
 
-    async def test_given_no_config_when_deploy_then_status_is_waiting(  # noqa: E501
-        self, ops_test, charm, cleanup
+    @pytest.mark.abort_on_fail
+    async def test_given_no_config_when_deploy_then_status_is_blocked(  # noqa: E501
+        self,
+        ops_test,
+        build_and_deploy,
```
Gmerold, line 112:
```suggestion
        self, ops_test, build_and_deploy
```

tests/integration/test_integration.py
```diff
@@ -133,17 +126,17 @@ async def test_given_no_config_when_post_deployment_tasks_and_authorise_charm_th
...
-
+        vault_token = post_deployment_tasks
         await vault_unit.run_action(action_name="authorise-charm", token=vault_token)
+
         await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
 
+    @pytest.mark.abort_on_fail
     async def test_given_status_is_active_when_run_issue_certificate_action_then_certificates_are_issued(  # noqa: E501
-        self, ops_test, charm, cleanup
+        self, ops_test, build_and_deploy, post_deployment_tasks
```
Gmerold, line 139: Update docstring to match new args. 
You can also drop the `Returns` section from it.

tests/integration/test_integration.py
Gmerold: Also should this be a fixture? To me, it looks like a private method.

tests/integration/test_integration.py
```diff
@@ -133,17 +126,17 @@ async def test_given_no_config_when_post_deployment_tasks_and_authorise_charm_th
         Returns:
             None
         """
-        await self.deploy_charm(ops_test, charm)
         vault_unit = ops_test.model.units["vault-k8s/0"]
         await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
```
Gmerold, line 130: I don't think we need this check. Previous test ensured this.

tests/integration/test_integration.py
```diff
@@ -114,15 +105,17 @@ async def post_deployment_tasks(self, namespace: str) -> str:
...
+        build_and_deploy,
     ):
-        await self.deploy_charm(ops_test, charm)
-
         await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
 
+    @pytest.mark.abort_on_fail
     async def test_given_no_config_when_post_deployment_tasks_and_authorise_charm_then_status_is_active(  # noqa: E501
-        self, ops_test, charm, cleanup
+        self, ops_test, build_and_deploy, post_deployment_tasks
```
Gmerold, line 118: Update docstring.

Gmerold (changes requested):

tests/integration/test_integration.py
sanchezfdezjavier: Docstring done. The reason I make it a fixture is for being run just once, I don't want to generate multiple tokens, is there a better way?

tests/integration/test_integration.py
sanchezfdezjavier: Black doesn't like it in the same line.

tests/integration/test_integration.py
sanchezfdezjavier: Both done.

tests/integration/test_integration.py
sanchezfdezjavier: True, removed.

tests/integration/test_integration.py
Gmerold: The trick is with the last comma ;) If you remove it, black will be happy.

tests/integration/test_integration.py
Gmerold: Now you run it at least twice. First time it's run as a fixture and the second is when you call it directly in the `test_given_no_config_when_post_deployment_tasks_and_authorise_charm_then_status_is_active` (btw, I think it should be authorize, with `z`, not `s`) test, to get the token.
You need post deployment tasks to get the token you'll use to authorize the charm. When you run post installs as a fixture, you're not even fetching this token.

tests/integration/test_integration.py
sanchezfdezjavier: Done.

tests/integration/test_integration.py
sanchezfdezjavier: - Am I running it twice? Correct me if I'm wrong, I think the point of `@pytest.fixture(scope="module")` is for that function to be run only once per module. Am I calling it here? `vault_token = post_deployment_tasks` is a reference to the fixture, so it's not being run there, and because the return value should be stored from the first and only run.
- Why am I not fetching the token, and how can that be possible if the tests succeed?
- I don't get this "(btw, I think it should be authorize, with z, not s)", the action in the `actions.yaml` is `authorise-charm`, I guess you don't mean that.

Gmerold: Damn, my bad, I don't why I saw `vault_token = post_deployment_tasks()`. Sorry.
Either way though, I'd rather make `post_deployment_tasks` a regular method. It's gonna be run only once anyway, because only one test uses it.
About the typo - yes, that's exactly what I meant. It should be `authorize` instead of `authorise`.

my_env/bin/python
```diff
@@ -0,0 +1 @@
+python3
```
gruyaume, line 1: What are those `my_env`?

tests/integration/test_integration.py
```diff
@@ -114,60 +103,48 @@ async def post_deployment_tasks(self, namespace: str) -> str:
...
+            post_deployment_tasks: Pytest fixture
         """
-        await self.deploy_charm(ops_test, charm)
-        vault_unit = ops_test.model.units["vault-k8s/0"]
-        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
-
-        vault_token = await self.post_deployment_tasks(namespace=ops_test.model_name)
+        vault_unit = ops_test.model.units["vault-k8s/0"]  # type: ignore[union-attr]
 
+        vault_token = post_deployment_tasks
```
gruyaume, line 125: Why remove the `await` here?

tests/integration/test_integration.py
sanchezfdezjavier: About the typo, that's not the difference between American and British English? The one you're suggesting is the American, doesn't it?

tests/integration/test_integration.py
sanchezfdezjavier: This is a fixture, I think we don't need to await.

tests/integration/test_integration.py
Gmerold: Pytest runs tests in the same order you put them in your test class. Always. This means, you don't need `post_deployment_tasks` to be a fixture. 

You're right about American vs British English :D

sanchezfdezjavier: And you're right about the fixture. Fixed :)

my_env/bin/python
sanchezfdezjavier: My bad, removed.

tests/integration/test_integration.py
sanchezfdezjavier: This comment it's deprecated, it's no longer a fixture.

tests/integration/test_integration.py
```diff
@@ -22,30 +22,13 @@
...
-    async def cleanup(self, ops_test):
-        try:
-            await ops_test.model.remove_application(
-                app_name=APPLICATION_NAME, block_until_done=True
-            )
-        except JujuError:
-            pass
-
     @staticmethod
     async def wait_for_load_balancer_address(kubernetes: Kubernetes, timeout: float = 60):
```
gruyaume, line 26: `timeout` should probably be an `int`. Sorry it's unrelated to your PR.

gruyaume, line 26: We should also add return typing

sanchezfdezjavier: Added. I've also added additional docstrings and fixed a return type.

tests/integration/test_integration.py
```diff
@@ -22,45 +22,45 @@
...
-            )
-        except JujuError:
-            pass
+        Args:
+            kubernetes: Kubernetes object.
+            timeout: Timeout (seconds).
 
-    @staticmethod
-    async def wait_for_load_balancer_address(kubernetes: Kubernetes, timeout: float = 60):
+        Retunrs:
```
Gmerold, line 35:
```suggestion
        Returns:
```

tests/integration/test_integration.py
```diff
@@ -22,45 +22,45 @@
...
         Args:
             vault: Vault object.
-            timeout: Timeout (seconds)
+            timeout: Timeout (seconds).
 
         Returns:
-            str: Vault's Unseal key
-            str: Vault's Root token
+            str: Vault's Unseal key.
+            str: Vault's Root token,
```
Gmerold, line 60:
```suggestion
            str: Vault's Root token.
```

tests/integration/test_integration.py
```diff
@@ -72,102 +72,100 @@ async def initialize_vault(vault: Vault, timeout: int = 60) -> Tuple[str, str]:
...
 
         Args:
-            ops_test: Ops test Framework
-            charm: Charm path
-
-        Returns:
-            None
+            ops_test: Ops test Framework.
+            build_and_deploy: Pytest fixture.
+            post_deployment_tasks: Pytest fixture.
```
Gmerold, line 142:
```suggestion
```

tests/integration/test_integration.py
```diff
@@ -72,102 +72,100 @@ async def initialize_vault(vault: Vault, timeout: int = 60) -> Tuple[str, str]:
...
```
Gmerold, line 160:
```suggestion
```

tests/integration/test_integration.py
```diff
@@ -72,102 +72,100 @@ async def initialize_vault(vault: Vault, timeout: int = 60) -> Tuple[str, str]:
...
-        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
-
-        vault_token = await self.post_deployment_tasks(namespace=ops_test.model_name)
+        vault_unit = ops_test.model.units["vault-k8s/0"]  # type: ignore[union-attr]
 
+        vault_token = await self.post_deployment_tasks(ops_test)
         await vault_unit.run_action(action_name="authorise-charm", token=vault_token)
-        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
 
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)  # type: ignore[union-attr]
```
Gmerold, line 149: Isn't lint going crazy about the line being too long?

tests/integration/test_integration.py
```diff
@@ -72,102 +72,100 @@ async def initialize_vault(vault: Vault, timeout: int = 60) -> Tuple[str, str]:
...
-    async def test_given_no_config_when_deploy_then_status_is_waiting(  # noqa: E501
-        self, ops_test, charm, cleanup
+    @pytest.mark.abort_on_fail
+    async def test_given_no_config_when_deploy_then_status_is_blocked(  # noqa: E501
+        self, ops_test: OpsTest, build_and_deploy
     ):
-        await self.deploy_charm(ops_test, charm)
-
-        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)  # type: ignore[union-attr]
```
Gmerold, line 131: Isn't lint going crazy about the line being too long?

tests/integration/test_integration.py
```diff
@@ -5,12 +5,12 @@
 import logging
 import time
 from pathlib import Path
-from typing import Tuple
+from typing import Optional, Tuple
 
 import pytest
 import requests.exceptions  # type: ignore[import]
```
Gmerold, line 11: You could add `types-requests` to the deps of the `static` env in the `tox.ini` and avoid this ignore.

tests/integration/test_integration.py
```diff
@@ -22,45 +22,45 @@
...
+            timeout: Timeout (seconds).
 
         Returns:
-            str: Vault's Unseal key
-            str: Vault's Root token
+            str: Vault's Unseal key.
+            str: Vault's Root token,
+
+        Raises:
+            TimeoutError: Whether Vault is not ready after timeout.
```
Gmerold, line 63: I might be wrong again, but I feel like `Whether` is good when you describe a boolean value. Here I'd rather use `When`. What do you think?

tests/integration/test_integration.py
```diff
@@ -22,45 +22,45 @@
...
+            kubernetes: Kubernetes object.
+            timeout: Timeout (seconds).
 
-    @staticmethod
-    async def wait_for_load_balancer_address(kubernetes: Kubernetes, timeout: float = 60):
+        Retunrs:
+            str: LoadBalancer address.
+
+        Raises:
+            TimeoutError: Whether LoadBalancer address is not available after timeout.
```
Gmerold, line 39: Same as below.

tests/integration/test_integration.py
```diff
@@ -72,102 +72,100 @@ async def initialize_vault(vault: Vault, timeout: int = 60) -> Tuple[str, str]:
...
+        """Builds and deploys vault-k8s charm.
+
+        Args:
+            ops_test: Ops test Framework.
+        """
+        ops_test.destructive_mode = False
+        charm = await ops_test.build_charm(".")
+        await self.deploy_charm(ops_test, charm)
+
+    @pytest.mark.abort_on_fail
```
Gmerold, line 105: I don't think you need this anymore. Test using this method is decorated with `abort_on_fail` anyway.

tests/integration/test_integration.py
sanchezfdezjavier: Fixed.

tests/integration/test_integration.py
sanchezfdezjavier: Fixed.

tests/integration/test_integration.py
sanchezfdezjavier: Removed.

tests/integration/test_integration.py
sanchezfdezjavier: Removed.

tests/integration/test_integration.py
sanchezfdezjavier: No, but I fixed it and it's not complaining.

tests/integration/test_integration.py
sanchezfdezjavier: No, but I fixed it and it's not complaining.

tests/integration/test_integration.py
sanchezfdezjavier: Done.

tests/integration/test_integration.py
sanchezfdezjavier: You're right, changed.

tests/integration/test_integration.py
sanchezfdezjavier: Fixed.

tests/integration/test_integration.py
sanchezfdezjavier: Why we don't? If that fails why would we want to continue? And sorry for the huge amount of typos, I went to fast.

Gmerold: `test_given_no_config_when_post_deployment_tasks_and_authorise_charm_then_status_is_active` uses this method. And it's gonna abort the run if it fails. Can the test succeed if post install fails?

sanchezfdezjavier: Yes, fixed.

Gmerold (approved):
LGTM

PR #31, opened 2023-08-21
[Feat] Uses clusterIP instead of LoadBalancer service

gruyaume:
# Description

Uses the default K8s service for charms (clusterIP) instead of a loadbalancer service. If LoadBalancer is required, this should be done by connecting to traffic instead.

# Checklist:


ghislainbourgeois (approved):
Since this is a breaking change, should we have a new track to let people upgrade on their own terms?

gruyaume:
> Since this is a breaking change, should we have a new track to let people upgrade on their own terms?

@ghislainbourgeois 
Right now this charm is only available on edge channel and doesn't support most of what we would consider production ready (HA + HTTPS).

In other words I'm not too worried about introducing breaking changes, at least not until we move to stable.

ghislainbourgeois:
> > Since this is a breaking change, should we have a new track to let people upgrade on their own terms?
> 
> @ghislainbourgeois 
> Right now this charm is only available on edge channel and doesn't support most of what we would consider production ready (HA + HTTPS).
> 
> In other words I'm not too worried about introducing breaking changes, at least not until we move to stable.

Ok, that works for me.

gboutry:
With this change, Vault won't be accessible from other nodes that don't have k8s networking defined, is that intended ?

PR #35, opened 2023-09-06
[Feat] Implements raft backend (HA)

gruyaume:
# Description

Implementation of HA leveraging the Raft backend. The leader initializes Vault and as unit gets added to the application, they look at the existing unit addresses from the peer relation data and use those in their configuration file to get added to the raft cluster.

## Reference
- https://developer.hashicorp.com/vault/docs/configuration/storage/raft

# Checklist:


src/charm.py
```diff
@@ -95,17 +113,62 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
             self.unit.status = WaitingStatus("Waiting for vault initialization secret")
             event.defer()
             return
+        if not self.unit.is_leader() and len(self._other_peer_unit_addresses()) == 0:
```
patriciareinoso, line 116: The spec states that the Vault leader may not be the same as the Juju leader. Should this be considered at this point?

Same question for the _on_install method.

gruyaume: By "may" not I mean that we don't control this, it could be it or not but that is decided through raft's concensus algorithm, not the charm.

src/charm.py
```diff
@@ -95,17 +113,62 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+            return
+        root_token, unseal_keys = self._get_initialization_secret_from_peer_relation()
+        if root_token:
+            vault = Vault(url=self._api_address)
+            vault.set_token(token=root_token)
+            if vault.is_api_available() and vault.node_in_raft_peers(node_id=self._node_id):
+                vault.remove_raft_node(node_id=self._node_id)
+        if self._vault_service_is_running():
+            self._container.stop(self._service_name)
+        self._container.remove_path(path=f"{VAULT_RAFT_DATA_PATH}/*", recursive=True)
```
ghislainbourgeois, line 156: Do we really need to remove the data? Couldn't this be left to the admin to remove the storage?

src/vault.py
```diff
@@ -88,3 +64,16 @@ def is_api_available(self) -> bool:
     def set_token(self, token: str) -> None:
         """Sets the Vault token for authentication."""
         self._client.token = token
+
+    def remove_raft_node(self, node_id: str) -> None:
+        """Remove raft peer."""
+        self._client.sys.remove_raft_node(server_id=node_id)
+        logger.info(f"Removed raft node {node_id}")
```
ghislainbourgeois, line 71:
```suggestion
        logger.info("Removed raft node %s", node_id)
```

tests/integration/test_integration.py
```diff
@@ -54,3 +55,27 @@ async def test_given_default_config_when_deploy_then_status_is_active(
...
+    async def test_given_application_is_deployed_when_scale_up_then_status_is_active(
+        self,
+        ops_test: OpsTest,
+        build_and_deploy,
+    ):
+        await ops_test.model.applications[APPLICATION_NAME].scale(7)  # type: ignore[union-attr]
+
+        await ops_test.model.wait_for_idle(  # type: ignore[union-attr]
+            apps=[APPLICATION_NAME], status="active", timeout=1000
+        )
```
ghislainbourgeois, line 69: You will also need to add the parameter `wait_for_exact_units` to the number you are expecting, otherwise this will always pass because one unit is already `active/idle`.

tests/integration/test_integration.py
```diff
@@ -54,3 +55,27 @@ async def test_given_default_config_when_deploy_then_status_is_active(
...
+    async def test_given_application_is_deployed_when_scale_down_then_status_is_active(
+        self,
+        ops_test: OpsTest,
+        build_and_deploy,
+    ):
+        await ops_test.model.applications[APPLICATION_NAME].scale(3)  # type: ignore[union-attr]
+
+        await ops_test.model.wait_for_idle(  # type: ignore[union-attr]
+            apps=[APPLICATION_NAME], status="active", timeout=1000
+        )
```
ghislainbourgeois, line 81: Same as above.

ghislainbourgeois (changes requested):

src/charm.py
gruyaume: If we don't and that a unit gets removed then re-added, this unit number will never spin up properly:

```
2023-09-11T22:23:14.994Z [vault] 2023-09-11T22:23:14.994Z [INFO]  core: vault is unsealed
2023-09-11T22:23:14.994Z [vault] 2023-09-11T22:23:14.994Z [INFO]  storage.raft: entering follower state: follower="Node at 10.1.182.60:8201 [Follower]" leader-address= leader-id=
2023-09-11T22:23:14.994Z [vault] 2023-09-11T22:23:14.994Z [INFO]  core: entering standby mode
2023-09-11T22:23:17.732Z [pebble] GET /v1/plan?format=yaml 122.652µs 200
2023-09-11T22:23:31.570Z [vault] 2023-09-11T22:23:31.570Z [WARN]  storage.raft: heartbeat timeout reached, not part of a stable configuration or a non-voter, not triggering a leader election
```

I'm not sure of the reason yet, I'm looking this up

gruyaume: From what I have seen in [this thread](https://github.com/hashicorp/vault/issues/8854), it seems normal practice to clean the data of the node before having it rejoin so that it shows up as a new node. So I'd probably keep the behaviour as is unless you see something else.

tests/integration/test_integration.py
gruyaume: Done

src/charm.py
```diff
@@ -149,8 +149,11 @@ def _on_remove(self, event: RemoveEvent):
         if root_token:
             vault = Vault(url=self._api_address)
             vault.set_token(token=root_token)
-            if vault.is_api_available() and vault.node_in_raft_peers(node_id=self._node_id):
-                vault.remove_raft_node(node_id=self._node_id)
+            try:
+                if vault.is_api_available() and vault.node_in_raft_peers(node_id=self._node_id):
+                    vault.remove_raft_node(node_id=self._node_id)
+            except (requests.exceptions.ConnectionError, requests.exceptions.TooManyRedirects):
+                pass
```
ghislainbourgeois, line 156: I am curious when those exceptions could raise. I would guess when it is the last unit?

Would it make sense to log something here?

ghislainbourgeois (commented):
I have a small question remaining

README.md
```diff
@@ -3,19 +3,16 @@
...
+the Raft backend.
 
 ## Usage
 
 ### Deploy
 
 Deploy the charm:
 ```bash
-juju deploy vault-k8s --trust
+juju deploy vault-k8s -n 5 --trust
```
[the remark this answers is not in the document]
gruyaume: It does not **need** to be deployed with an odd number of units, Vault will work just fine with 4 units as with 3. It's just typically makes more sense if one wants to maximise the fault tolerance and minimise the number of nodes deployed.
- 1 node supports 0 failure (majority is 1 node)
- 2 nodes supports 0 failure (majority is 2 nodes)
- 3 nodes supports 1 failure (majority is 2 nodes)
- 4 nodes supports 1 failure (majority is 3 nodes)
- 5 nodes supports 2 failures (majority is 3 nodes)
- 6 nodes supports 2 failures (majority is 4 nodes)

src/charm.py
gruyaume: I found why we were getting the `TooManyRedirectErrors` (the bind_address) is not always available on remove event. This  try/except is not necessary anymore and it's been removed

src/vault.py
```diff
@@ -88,3 +64,16 @@ def is_api_available(self) -> bool:
...
+
+    def remove_raft_node(self, node_id: str) -> None:
+        """Remove raft peer."""
+        self._client.sys.remove_raft_node(server_id=node_id)
+        logger.info("Removed raft node %s", node_id)
+
+    def node_in_raft_peers(self, node_id: str) -> bool:
+        """Check if node is in raft peers."""
+        raft_config = self._client.sys.read_raft_config()
+        for peer in raft_config["data"]["config"]["servers"]:
```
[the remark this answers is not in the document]
gruyaume: I haven't seen usecase where it wasn't but we can always add a validation when and if that occurs

ghislainbourgeois (approved):

tests/unit/test_vault.py
```diff
@@ -95,56 +41,64 @@ def test_given_n_unseal_keys_when_unseal_then_unseal_called_n_times(
...
-    def test_given_api_not_available_when_wait_for_api_available_then_timeouterror(
-        self,
-        patch_is_api_available,
+    @patch("hvac.api.system_backend.raft.Raft.read_raft_config")
+    def test_given_node_not_in_peer_list_when_is_node_in_raft_peers_then_returns_false(
+        self, patch_health_status
     ):
-        patch_is_api_available.return_value = False
+        node_id = "whatever node id"
         vault = Vault(url="http://whatever-url")
```
gatici, line 77: Just a minor thing, Vault URL could be stored in a variable. "http://whatever-url" is set multiple times.

gatici (approved):

tests/unit/test_vault.py
gruyaume: I think it's ok that it's initialised multiple times, it makes each test readable. I'd rather optimise for readability then efficiency

PR #36, opened 2023-09-13
[feat] Enables TLS in Vault cluster

gruyaume:
# Description

Here we enable TLS in the Vault cluster:
- On install, a private key, CA certificate and certificate are generated, stored in juju secret and pushed to the workload
- Those certificates are refered to by the nodes in their pebble configuration
- Modifications to README.md to display TLS related changes

_Note: Currently, we explicitely do not verify the CA in the vault API calls (`verify=False`). In a future PR we would probably want to enable this and reference to Vault's CA certificate. This PR is big and complex enough as is to include this change here._

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -271,7 +271,6 @@ class VaultCharm(CharmBase):
     ) -> bool:
         """Check if the certificates are pushed to the workload."""
         if not self._container.exists(path=TLS_CERT_FILE_PATH):
-            print("HELLO")
             return False
         if not self._container.exists(path=TLS_KEY_FILE_PATH):
             return False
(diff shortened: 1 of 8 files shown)


src/charm.py
```diff
@@ -188,9 +254,91 @@ def _vault_service_is_running(self) -> bool:
...
+            "vault-certificates-secret-id"
+        )
+        if not juju_secret_id:
+            return None, None, None
+        try:
+            juju_secret = self.model.get_secret(id=juju_secret_id)
+        except SecretNotFoundError:
+            return None, None, None
+        content = juju_secret.get_content()
+        return content["privatekey"], content["certificate"], content["cacertificate"]
```
ghislainbourgeois, line 341: I think this could be simplified to simply raise in error cases, and have an `except` in the event handler to set the status to waiting, it would get rid of a lot of conditionals.

ghislainbourgeois (changes requested):

src/charm.py
gruyaume: Done, let me know what you think

ghislainbourgeois (approved):

PR #39, opened 2023-09-19
[Fix] Add subject and ingress address to subject alt names

### LINKED ISSUE #37: [BUG] vault certificate only contains unit's IP address as SAN
**Describe the bug**
Vault server's certificate only defines the unit's IP address as a SAN

**To Reproduce**
1. Get vault peer secret storing certficates
2. Run `juju show-secret <secret-id> --reveal | yq '.<secret-id>.content.certificate' > server.crt`
3. Run `openssl x509 -in vault.crt -text -noout`
4. Look at the Subjet Alternative names

**Expected behavior**
Service being usable through the SVC address or SVC url.

**Screenshots**

**Logs**
In barbican error logs:
```
barbican.plugin.interface.secret_store.SecretGeneralException: Problem seen during crypto processing - Reason: HTTPSConnectionPool(host='vault.barbican.svc.cluster.local', port=8200): Max retries exceeded with url: /v1/auth/approle/login (Caused by SSLError(CertificateError("hostname 'vault.barbican.svc.cluster.local' doesn't match '10.1.166.227'")))
```

**Environment**
 - Vault version: 1.13.3
 - Juju version: 3.2.0
 - Cloud Environment: microk8s
 - Kubernetes version: 1.26

**Additional context**
I'm not sure what's the preferred way to contact Vault, should it be behind the K8S SVC ? (so either Cluster IP or URL), or should we form a quorum of url with each unit IP address ? In the first case, either the Cluster IP or (inclusive) the SVC URL should be added to the certificate. In the second case, you can close this bug, and I will use the unit's IP addresses.

gboutry:
# Description

Fix #37 

Services reaching Vault from inside K8S (no external addressing) use the loadbalancer primitives offered by K8S.

Adding to Subject Alt Names:
- Service name as DNS SAN
- Ingress address (ClusterIP) as IP SAN

# Checklist:

- [X] I validated that new and existing unit tests pass locally with my changes

# Notes

I'm not sure that's the right way to handle, since current TLS handling does not handle scaling, TLS management will most likely evolve. Happy to discard if that's not wished.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -60,7 +60,7 @@ class PeerSecretError(Exception):
         super().__init__(self.message)
 
 
-def generate_vault_certificates(subject: str, sans_ip: str) -> Tuple[str, str, str]:
+def generate_vault_certificates(subject: str, sans_ip: List[str]) -> Tuple[str, str, str]:
     """Generate Vault certificates valid for 50 years.
 
     Returns:
@@ -73,7 +73,9 @@ def generate_vault_certificates(subject: str, sans_ip: str) -> Tuple[str, str, s
         validity=365 * 50,
     )
     vault_private_key = generate_private_key()
-    csr = generate_csr(private_key=vault_private_key, subject=subject, sans_ip=[sans_ip])
+    csr = generate_csr(
+        private_key=vault_private_key, subject=subject, sans_ip=sans_ip, sans_dns=[subject]
+    )
     vault_certificate = generate_certificate(
         ca=ca_certificate,
         ca_key=ca_private_key,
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -430,6 +434,24 @@ def _bind_address(self) -> Optional[str]:
         except ModelError:
             return None
 
+    @property
+    def _ingress_address(self) -> Optional[str]:
```
gruyaume, line 438: We will soon add integration with an ingress (traefik) through the [ingress](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/ingress/v2) charm relation interface. Will we have 2 meanings for "ingress" or is it the same meaning but this `_ingress_address` returns something else when it's related to an ingress?

gboutry: Indeed, it can be confusing.

In theory, we have one meaning of ingress:

The case of wanting an unique URL to contact the Vault service (not a Vault unit).

Which is served by either the Juju's concept of ingress address or Traefik's ingress.


I believe it's fine to have the ingress property return either the traefik's ingress address when related to an ingress or return juju's concept of ingress address when it's missing.

:warning: 
This just means that in case of CMR that's cross cloud (Vault hosted on a k8s and requiring on any other cloud including an other k8s), the relation cannot work without traefik's ingress. Which is fine if we can alert that we're in a CMR (which I'm not sure is possible) or it's explicitly documented.

gboutry: Not sure how to handle it from a TLS perspective. It depends if traefik is configured in passthrough or reencrypt, but this should be determined when integrating with traefik.

gruyaume (approved):

PR #49, opened 2023-09-30
[bugfix] Fixes retry join comparison in config file

gruyaume:
# Description

Fixes retry join comparison in config file. The retryjoin stanza is typically a list. When comparing the file content, we should not care about ordering, therefore we are now converting to set before making the comparison.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -89,6 +89,43 @@ def render_vault_config_file(
     return content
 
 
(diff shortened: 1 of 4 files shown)


saltiyazan (approved):
Not strictly related to this change, but we don't validate the behavior of generating config files in tests.

And could the same issue happen with other config entries?

Otherwise looks good.

PR #51, opened 2023-10-01
[bugfix] Implements per-unit TLS certificate

### LINKED ISSUE #50: [BUG] Vault certificate is set for a specific unit IP but used by all units
**Describe the bug**

The generated certificate for vault is set for a specific unit IP address but we use the same cert for all the units. This leads to SSL errors when making calls to the vault unit that is not the one the cert was generated for. We probably will want to have 1 cert per unit.

**Expected behavior**

**Logs**

```
unit-vault-k8s-1: 18:13:29 ERROR unit.vault-k8s/1.juju-log vault-peers:13: Vault is not available: HTTPSConnectionPool(host='10.1.182.15', port=8200): Max retries exceeded with url: /v1/sys/health (Caused by SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: IP address mismatch, certificate is not valid for '10.1.182.15'. (_ssl.c:1007)")))
```

gruyaume:
# Description

Units used to share the same certificate which was causing an issue where SSL clients like the vault client would throw errors because the Vault instance IP did not match its certificate IP for non-leader units (since it was the leader that generated the cert). This PR makes it so that each unit generates its own certificate on charm install. They are all signed by the same CA certificate which is generated once by the leader.

Fixes #50 

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -137,11 +137,11 @@ class PeerSecretError(Exception):
         super().__init__(self.message)
 
 
-def generate_vault_certificates(subject: str, sans_ip: List[str]) -> Tuple[str, str, str]:
-    """Generate Vault certificates valid for 50 years.
+def generate_vault_ca_certificate() -> Tuple[str, str]:
+    """Generate Vault CA certificates valid for 50 years.
 
     Returns:
-        Tuple[str, str, str]: Private key, certificate, CA certificate
+        Tuple[str, str]: CA Private key, CA certificate
     """
     ca_private_key = generate_private_key()
     ca_certificate = generate_ca(
@@ -149,6 +149,23 @@ def generate_vault_certificates(subject: str, sans_ip: List[str]) -> Tuple[str, 
         subject="Vault self signed CA",
         validity=365 * 50,
     )
+    return ca_private_key.decode(), ca_certificate.decode()
+
+
+def generate_vault_unit_certificate(
+    subject: str, sans_ip: List[str], ca_certificate: bytes, ca_private_key: bytes
+) -> Tuple[str, str]:
+    """Generate Vault unit certificates valid for 50 years.
+
+    Args:
+        subject: Subject of the certificate
+        sans_ip: List of IP addresses to add to the SAN
+        ca_certificate: CA certificate
+        ca_private_key: CA private key
+
+    Returns:
+        Tuple[str, str]: Unit private key, Unit certificate
+    """
     vault_private_key = generate_private_key()
     csr = generate_csr(
         private_key=vault_private_key, subject=subject, sans_ip=sans_ip, sans_dns=[subject]
@@ -159,7 +176,7 @@ def generate_vault_certificates(subject: str, sans_ip: List[str]) -> Tuple[str, 
         csr=csr,
         validity=365 * 50,
     )
-    return vault_private_key.decode(), vault_certificate.decode(), ca_certificate.decode()
+    return vault_private_key.decode(), vault_certificate.decode()
 
 
 class VaultCharm(CharmBase):
@@ -204,12 +221,26 @@ class VaultCharm(CharmBase):
 
         Sets pebble plan, initializes vault, and unseals vault.
         """
+        if self.unit.is_leader():
+            self._on_install_leader(event)
+        else:
+            self._on_install_non_leader(event)
+
+    def _on_install_leader(self, event: InstallEvent):
+        """Install event handler for leader unit.
+
+        This handler is responsible for:
+        - Deleting pre-existing Vault data
+        - Creating CA certificate
+        - Creating leader unit certificate
+        - Initializing vault
+
+        Args:
+            event: InstallEvent
+        """
         if not self._container.can_connect():
             self.unit.status = WaitingStatus("Waiting to be able to connect to vault unit")
             event.defer()
-            return
-        self._delete_vault_data()
-        if not self.unit.is_leader():
             return
         if not self._is_peer_relation_created():
             self.unit.status = WaitingStatus("Waiting for peer relation to be created")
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -576,38 +636,26 @@ def _api_address(self) -> str:
...
 
-    def _push_certificates_to_workload(
-        self,
-        private_key: str,
-        certificate: str,
-        ca_certificate: str,
-    ) -> None:
-        """Push the certificates to the workload."""
-        self._container.push(path=TLS_CERT_FILE_PATH, source=certificate)
+    def _push_ca_certificate_to_workload(self, certificate: str) -> None:
```
gatici, line 639: Add Args in Docstring

src/charm.py
```diff
@@ -576,38 +636,26 @@ def _api_address(self) -> str:
...
+        """Push the CA certificate to the workload."""
+        self._container.push(path=TLS_CA_FILE_PATH, source=certificate)
+        logger.info("Pushed CA certificate to workload")
+
+    def _push_unit_certificate_to_workload(self, private_key: str, certificate: str) -> None:
```
gatici, line 644: Add Args, Returns in Docstring

src/charm.py
```diff
@@ -576,38 +636,26 @@ def _api_address(self) -> str:
...
-        if existing_private_key.read() != private_key:
-            return False
-        existing_ca_certificate = self._container.pull(path=TLS_CA_FILE_PATH)
-        if existing_ca_certificate.read() != ca_certificate:
-            return False
-        return True
+        self._container.push(path=TLS_CERT_FILE_PATH, source=certificate)
+        logger.info("Pushed unit certificate to workload")
+
+    def _ca_certificate_pushed_to_workload(self) -> bool:
```
gatici, line 650: Docstring arrangement as above

src/charm.py
```diff
@@ -576,38 +636,26 @@ def _api_address(self) -> str:
...
+        """Check if the CA certificate is pushed to the workload."""
+        return self._container.exists(path=TLS_CA_FILE_PATH)
+
+    def _unit_certificate_pushed_to_workload(self) -> bool:
```
gatici, line 654: Docstring arrangement as above

tests/unit/test_charm.py
```diff
@@ -132,19 +131,124 @@ def _set_other_node_api_address_in_peer_relation(self, relation_id: int, unit_na
...
+        patch_generate_unit_cert,
+        patch_get_binding,
+        patch_push,
+    ):
+        ca_certificate = "ca certificate"
+        ca_private_key = "ca private key"
+        unit_private_key = "unit private key"
+        unit_certificate = "unit certificate"
+        bind_address = "1.2.1.2"
+        ingress_address = "10.1.0.1"
```
gatici, line 179: Those parameters could be defined as global as they are defined several times in the tests.

gatici (commented):
Overall looks good. Minor comments are added.

src/charm.py
```diff
@@ -643,50 +691,44 @@ def _push_config_file_to_workload(self, content: str):
...
             "privatekey": private_key,
             "certificate": certificate,
-            "cacertificate": ca_certificate,
         }
-        juju_secret = self.app.add_secret(juju_secret_content, label="vault-certificate")
+        juju_secret = self.app.add_secret(juju_secret_content, label="vault-ca-certificate")
         peer_relation = self.model.get_relation(PEER_RELATION_NAME)
-        peer_relation.data[self.app].update({"vault-certificates-secret-id": juju_secret.id})  # type: ignore[union-attr]  # noqa: E501
-        logger.info("Vault certificate secret set in peer relation")
+        peer_relation.data[self.app].update({"vault-ca-certificates-secret-id": juju_secret.id})  # type: ignore[union-attr]  # noqa: E501
```
gatici, line 713: Suggest to define labels and secret-id names ("vault-ca-certificates-secret-id", "vault-ca-certificate" ..etc) as global variables

gatici (changes requested):

src/charm.py
```diff
@@ -222,31 +253,30 @@ def _on_install(self, event: InstallEvent):
...
-            ca_certificate=ca_certificate,
-        ):
-            self._push_certificates_to_workload(
-                certificate=certificate,
-                private_key=private_key,
-                ca_certificate=ca_certificate,
+            self._push_unit_certificate_to_workload(
+                certificate=certificate, private_key=private_key
             )
+        self._delete_vault_data()
```
patriciareinoso, line 281: why change the order of execution and delete the vault data after dealing with the certificates?

src/charm.py
gruyaume: Done

src/charm.py
gruyaume: Done

src/charm.py
gruyaume: I improved the docstring but ultimately I don't agree that we should add return docstring in such obvious cases. That would be just for the sake of it and not make the code more readable.

src/charm.py
gruyaume: I improved the docstring but ultimately I don't agree that we should add return docstring in such obvious cases. That would be just for the sake of it and not make the code more readable.

tests/unit/test_charm.py
gruyaume: I disagree, each of those is used in the scope of their own test. It's true that we ended up using the same values in different tests but they are not "global"

src/charm.py
gruyaume: Done

src/charm.py
gruyaume: I placed everything that could end up in deferring before so that we don't call it for no reason.

gatici (approved):

PR #55, opened 2023-10-03
Simplify test by using Harness get_filesystem_root and add_network

benhoyt:
Per discussion with Guillaume, we want to reduce the number of mocks in tests. In particularly, here we can use:

- Harness.get_filesystem_root method to make to make the push/pull operations more real and avoid patching Container push/pull/exists
- Harness.add_network to avoid patching Model.get_binding

In addition, tighten up the test a fair bit. Instead of exists always returning True, create the specific files we're looking for.

Also assert that vault.unseal() was actually called. It was too easy to break the test and accidentally follow the wrong code path, and the test still succeeded. For example, if you removed the patch_get_binding before, the test would still pass fine, but would be following one of the early returns -- not what the test was trying to test for.

Also pin ops to a version compatible with the latest version to ensure we have the newish features.

Not saying you have to merge this as is, but this is an example of how to 1) tighten up assertions, and 2) use the newish filesystem and network fakes in Harness.

# Checklist:

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -498,50 +498,40 @@ class TestCharm(unittest.TestCase):
             ]
         )
 
-    @patch("vault.Vault.unseal", new=Mock)
+    @patch("vault.Vault.unseal")
     @patch("vault.Vault.initialize")
     @patch("vault.Vault.is_api_available")
-    @patch("ops.model.Container.exists")
-    @patch("ops.model.Container.pull")
-    @patch("ops.model.Container.push")
-    @patch("ops.model.Model.get_binding")
-    @patch("charm.generate_vault_unit_certificate")
-    @patch("charm.generate_vault_ca_certificate")
     def test_given_certificates_pushed_when_install_then_certificate_not_pushed_to_workload(
         self,
-        patch_generate_ca_certs,
-        patch_generate_unit_certs,
-        patch_get_binding,
-        patch_push,
-        patch_pull,
-        patch_exists,
         patch_is_api_available,
         patch_vault_initialize,
+        patch_vault_unseal,
     ):
-        ca_private_key = "ca private key content"
-        ca_certificate = "ca certificate content"
-        unit_private_key = "unit private key content"
-        unit_certificate = "unit certificate content"
-
-        patch_exists.return_value = True
-        patch_pull.return_value = StringIO()
-        patch_generate_ca_certs.return_value = ca_private_key, ca_certificate
-        patch_generate_unit_certs.return_value = unit_private_key, unit_certificate
-        bind_address = "1.2.3.4"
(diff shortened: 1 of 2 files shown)


benhoyt:
Oops, the test I actually intended to fix here was the one Guillaume linked me to, [`test_given_certificates_pushed_when_install_then_certificate_not_pushed_to_workload`](https://github.com/canonical/vault-k8s-operator/blob/da4a7e2ae1fe9d4d26f5e622737fabe62626e7e8/tests/unit/test_charm.py#L420). Oh well, hopefully this one provides an example you can replicate. :-)

tests/unit/test_charm.py
```diff
@@ -498,50 +498,40 @@ def test_given_can_connect_when_install_then_ca_certificate_pushed_to_workload(
...
 
-        for call_args in patch_push.call_args_list:
-            self.assertFalse(call_args.kwargs["path"].startswith("/vault/certs"))
+        # Ensure CA and key+cert weren't overwritten
+        self.assertEqual((root / "vault/certs/ca.pem").read_text(), "CA")
+        self.assertEqual((root / "vault/certs/key.pem").read_text(), "KEY")
+        self.assertEqual((root / "vault/certs/cert.pem").read_text(), "CERT")
+
+        # Ensure vault was actually unsealed
+        patch_vault_unseal.assert_called_once()
```
gruyaume, line 534: I'm not sure why we added this here, that's not what the test name claims to validate

gruyaume (commented):
Overall I really like this approach, this will clean tests up

tests/unit/test_charm.py
benhoyt: See my comment in the PR description:

> Also assert that vault.unseal() was actually called. It was too easy to break the test and accidentally follow the wrong code path, and the test still succeeded. For example, if you removed the patch_get_binding before, the test would still pass fine, but would be following one of the early returns -- not what the test was trying to test for.

I think it sort of is what the test is trying to validate (previously the charm event could return early, not follow the code path we were looking for at all, but still pass. We're checking that certificates are *not* pushed (something did *not* happen), so we have to validate that it actually got up to the certificate check, rather than returned before then.

This might not be the right assertion though. Feel free to suggest something else.

Could you guys please take this from here, as you know the charm a lot better? (And can resolve the conflicts. :-)

gruyaume: Yes, I've created a new PR to apply this proposal all over the place:
- https://github.com/canonical/vault-k8s-operator/pull/59

gruyaume:
Closing and taking this change part of PR #59

tests/unit/test_charm.py
benhoyt: Excellent, thanks!

PR #61, opened 2023-10-12
feat: Enables audit device upon init

saltiyazan:
# Description

Enables audit device upon initialisation of Vault

Adds unit tests.
Fixes tests.
Fixes tox issues.
Bumps lib patch to publish change from [this merged pr](https://github.com/canonical/vault-k8s-operator/commit/c5a7608a3619f739ac6449aea39e60a40c32ca3a#diff-e441c2125fb9fd9548dec33cc23c5d2cf968aa96e7b516a2c5f1045148470593R123).

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_kv.py
+++ b/lib/charms/vault_k8s/v0/vault_kv.py
@@ -134,7 +134,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 PYDEPS = ["pydantic", "pytest-interface-tester"]
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -401,6 +401,7 @@ class VaultCharm(CharmBase):
             return
         if vault.is_sealed():
             vault.unseal(unseal_keys=unseal_keys)
+        vault.wait_for_unseal()
         vault.enable_audit_device(device_type="file", path="stdout")
         self._set_peer_relation_node_api_address()
         self.unit.status = ActiveStatus()
--- a/src/charm.py
+++ b/src/charm.py
@@ -237,7 +237,7 @@ class VaultCharm(CharmBase):
     def _on_install(self, event: InstallEvent):
         """Handler triggered when the charm is installed.
 
-        Sets pebble plan, initializes vault, and unseals vault.
+        Sets pebble plan, initializes vault, enable audit device, and unseals vault.
         """
         if self.unit.is_leader():
             self._on_install_leader(event)
@@ -401,6 +401,7 @@ class VaultCharm(CharmBase):
             return
         if vault.is_sealed():
             vault.unseal(unseal_keys=unseal_keys)
+        vault.enable_audit_device(device_type="file", path="stdout")
         self._set_peer_relation_node_api_address()
         self.unit.status = ActiveStatus()
 
--- a/src/vault.py
+++ b/src/vault.py
@@ -104,6 +104,15 @@ class Vault:
             mount_policy = fd.read()
         self._client.sys.create_or_update_policy(policy, mount_policy.format(mount=mount))
 
+    def enable_audit_device(self, device_type: str, path: str) -> None:
+        """Enable a new audit device at the supplied path."""
+        if device_type + "/" not in self._client.sys.list_enabled_audit_devices()["data"].keys():
+            self._client.sys.enable_audit_device(
+                device_type=device_type,
+                options={"file_path": path},
+            )
+            logger.info("Enabled audit device of type: %s, using path: %s", device_type, path)
+
     def configure_approle(self, name: str, cidrs: List[str], policies: List[str]) -> str:
         """Create/update a role within vault associating the supplied policies."""
         self._client.auth.approle.create_or_update_approle(
--- a/src/vault.py
+++ b/src/vault.py
@@ -48,21 +48,21 @@ class Vault:
         """Returns whether Vault is sealed."""
         return self._client.sys.is_sealed()
 
-    def wait_for_unseal(self) -> None:
+    def wait_for_unseal(self, max_attempts: int = 15) -> None:
         """Waits for Vault to be unsealed.
 
         Expected to be called after attempting to unseal Vault.
         If it times out, raises a TimeoutError.
+
+        Args:
+            max_attempts: The maximum number of attempts to check if Vault is unsealed.
         """
-        elapsed_time = 0
-        timeout = 30
-        while elapsed_time < 30:
+        for _ in range(max_attempts):
             if not self.is_sealed():
                 return
             logger.info("Vault is sealed, waiting for unseal")
             time.sleep(2)
-            elapsed_time += 2
-        raise TimeoutError(f"Vault is still sealed after waiting for {timeout} seconds")
+        raise TimeoutError(f"Vault is still sealed after checking {max_attempts} times")
 
     def unseal(self, unseal_keys: List[str]) -> None:
         """Unseal Vault."""
--- a/src/vault.py
+++ b/src/vault.py
@@ -63,6 +63,11 @@ class Vault:
                 return
             logger.info("Vault is sealed, waiting for unseal")
             time.sleep(2)
+
+        # One more check after the loop to catch any edge cases.
+        if not self.is_sealed():
+            return
+
         raise TimeoutError(f"Vault is still sealed {timeout} seconds")
 
     def unseal(self, unseal_keys: List[str]) -> None:
@@ -136,6 +141,7 @@ class Vault:
                     "Failed to enable audit device of type: %s, using path: %s", device_type, path
                 )
             logger.info("Enabled audit device of type: %s, using path: %s", device_type, path)
+        logger.info("Audit device of type: %s, using path: %s already enabled", device_type, path)
 
     def configure_approle(self, name: str, cidrs: List[str], policies: List[str]) -> str:
         """Create/update a role within vault associating the supplied policies."""
--- a/src/vault.py
+++ b/src/vault.py
@@ -48,21 +48,22 @@ class Vault:
         """Returns whether Vault is sealed."""
         return self._client.sys.is_sealed()
 
-    def wait_for_unseal(self, max_attempts: int = 15) -> None:
+    def wait_for_unseal(self, timeout: int = 30) -> None:
         """Waits for Vault to be unsealed.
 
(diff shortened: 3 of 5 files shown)


tests/unit/test_charm.py
gruyaume, line 1: We should add a test to make sure we call `vault.enable_audit_device` on install event.

src/charm.py
```diff
@@ -401,6 +401,7 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
             return
         if vault.is_sealed():
             vault.unseal(unseal_keys=unseal_keys)
+        vault.enable_audit_device(device_type="file", path="stdout")
```
gruyaume, line 404: Looking at the integration tests run, it looks like we run this too quickly after the unseal. there may be a validation to add here.

tests/unit/test_charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/vault.py
```diff
@@ -47,6 +48,15 @@ def is_sealed(self) -> bool:
         """Returns whether Vault is sealed."""
         return self._client.sys.is_sealed()
 
+    def wait_for_unseal(self) -> None:
+        """Wait for Vault to be unsealed."""
```
DanielArndt, line 52: This waits for Vault to be unsealed, up to a timeout. If it takes longer than that, we proceed anyway.

I'm not sure if that is intentional, but the docstring should probably make that clear. Perhaps we want to throw an exception if the Vault is still sealed after the timeout? Or log a warning? I'm not sure of the wider context and what would be more appropriate.

src/vault.py
```diff
@@ -47,6 +48,15 @@ def is_sealed(self) -> bool:
...
+        elapsed_time = 0
+        while self.is_sealed() and elapsed_time < 30:
+            logger.info("Vault is sealed, waiting for unseal")
+            time.sleep(5)
```
DanielArndt, line 56: 5 seconds seems like a long time to wait between polls.

src/vault.py
saltiyazan: Good point.
I think throwing an exception makes sense, as continuing with a sealed vault and then trying to create the audit device would throw an exception too.

Done.

src/vault.py
saltiyazan: Reduced to 2.

src/vault.py
saltiyazan: Do you think it would be a better approach if we call `vault.unseal` in `wait_for_unseal`? or should we keep it as it is now?

src/vault.py
```diff
@@ -47,6 +48,22 @@ def is_sealed(self) -> bool:
...
 
+    def wait_for_unseal(self) -> None:
+        """Waits for Vault to be unsealed.
+
+        Expected to be called after attempting to unseal Vault.
+        If it times out, raises a TimeoutError.
+        """
+        elapsed_time = 0
+        timeout = 30
+        while elapsed_time < 30:
```
gruyaume, line 59: We've just defined a "timeout" varaible, might as well use it :)

gruyaume (changes requested):

src/vault.py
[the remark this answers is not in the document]
saltiyazan: This would be a bit hard to test/mock
I changed it in a different way.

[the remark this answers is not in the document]
gruyaume: I don't think we should settle for a worse design becaues it's "a bit hard to test". Here how the tests could look like:

```python

import unittest
from unittest.mock import patch, Mock


from vault import Vault


def infinite_time_values(start=0, step=2):
    """Generator that returns an infinite sequence of time values.

    Args:
        start: Initial time value.
        step: Time step between values.
    """
    current_time = start
    while True:
        yield current_time
        current_time += step


class TestVault(unittest.TestCase):

    @patch("vault.Vault.is_sealed")
    @patch("time.sleep", new=Mock)
    @patch("time.time")
    def test_given_vault_is_sealed_when_wait_for_unseal_then_timeout_error(
        self, mock_time, mock_is_sealed
    ):
        mock_is_sealed.return_value = True

        time_values = infinite_time_values()

        mock_time.side_effect = lambda: next(time_values)

        vault = Vault(url="http://whatever-url", ca_cert_path="whatever path")

        with self.assertRaises(TimeoutError):
            vault.wait_for_unseal(timeout=30)

    @patch("vault.Vault.is_sealed")
    @patch("time.sleep", new=Mock)
    def test_given_vault_is_unsealed_when_wait_for_unseal_then_returns(
        self, mock_is_sealed
    ):
        mock_is_sealed.return_value = False

        vault = Vault(url="http://whatever-url", ca_cert_path="whatever path")

        vault.wait_for_unseal(timeout=30)
```

And the associated method:

```python
    def wait_for_unseal(self, timeout: int = 30) -> None:
        """Waits for Vault to be unsealed.

        Expected to be called after attempting to unseal Vault.
        If it times out, raises a TimeoutError.

        Args:
            timeout: Timeout in seconds.
        """
        initial_time = time.time()
        while time.time() - initial_time < timeout:
            if not self.is_sealed():
                return
            logger.info("Vault is sealed, waiting for unseal")
            time.sleep(2)
        raise TimeoutError(f"Vault is still sealed {timeout} seconds")
```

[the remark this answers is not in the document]
saltiyazan: I changed it, although I don't really see a clear reason why the other design is a bad one. And by hard to test I didn't mean difficult, but more complex and adds one more patch to the already long list of patches we have :)

gruyaume (approved):

src/vault.py
```diff
@@ -48,14 +48,22 @@ def is_sealed(self) -> bool:
...
+        """
+        initial_time = time.time()
+        while time.time() - initial_time < timeout:
+            if not self.is_sealed():
+                return
             logger.info("Vault is sealed, waiting for unseal")
-            time.sleep(5)
-            elapsed_time += 5
-        return
+            time.sleep(2)
```
DanielArndt, line 65: I like this pattern more (using time instead of counting). But there's a catch here:

Thinking about worst case scenario, pretend the scheduler doesn't get around to waking the process back up after 2 seconds. It's really busy and it takes 31 seconds. We will check the vault once right at the beginning, and sleep for 30 seconds, then raise a `TimeoutError`. Obviously this is an extreme situation to highlight a point, but it makes the problem clear I hope. We're waiting _at least_ 2 seconds in the end, for nothing (no additional work will be done afterwards).

Two ways to work around it come to mind:

1. Check if it's sealed before entering the loop, and sleep at the start of the loop instead of the end
2. Do one last check before you throw the error.

tests/unit/test_charm.py
```diff
@@ -19,6 +19,19 @@
...
+    """Generator that returns an infinite sequence of time values.
+
+    Args:
+        start: Initial time value.
+        step: Time step between values.
+    """
+    current_time = start
+    while True:
+        yield current_time
+        current_time += step
```
DanielArndt, line 32: You could use `itertools.count()` for this

```
>>> from itertools import count
>>> a = count(0,2)
>>> next(a)
0
>>> next(a)
2
>>> next(a)
4
>>> next(a)
6
>>> next(a)
8
```

DanielArndt (changes requested):
A couple of minor comments but otherwise LGTM

src/vault.py
```diff
@@ -104,6 +122,21 @@ def configure_kv_policy(self, policy: str, mount: str):
...
+                self._client.sys.enable_audit_device(
+                    device_type=device_type,
+                    options={"file_path": path},
+                ).status_code
+                != 204
+            ):
+                logger.error(
+                    "Failed to enable audit device of type: %s, using path: %s", device_type, path
+                )
+            logger.info("Enabled audit device of type: %s, using path: %s", device_type, path)
```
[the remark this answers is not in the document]
saltiyazan: Added it for now, but I'm hesitant about this, we will have this logged every time `on_config_changed` runs, and the user is not explicitly trying to enable the audit device. Wouldn't it make the logs a bit noisy?

tests/unit/test_charm.py
saltiyazan: Nice ✅

src/vault.py
saltiyazan: Good catch, I added one more check.

src/vault.py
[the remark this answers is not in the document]
gatici: I think it's fine to log that audit device is enabled as info level.

src/vault.py
gatici: IMO,  If a single method includes all the logic for unsealing including the wait for timeout, it would be better.

src/vault.py
```diff
@@ -104,6 +127,22 @@ def configure_kv_policy(self, policy: str, mount: str):
             mount_policy = fd.read()
         self._client.sys.create_or_update_policy(policy, mount_policy.format(mount=mount))
 
+    def enable_audit_device(self, device_type: str, path: str) -> None:
+        """Enable a new audit device at the supplied path."""
```
gatici, line 131: We may add Args to docstring

gatici:
Overall looks good, some minor comments are added.

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -134,7 +134,7 @@ def get_nonce(self):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
```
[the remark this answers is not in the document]
saltiyazan: Code was changed in a previous pr and that was not published.
Here: https://github.com/canonical/vault-k8s-operator/commit/c5a7608a3619f739ac6449aea39e60a40c32ca3a#diff-e441c2125fb9fd9548dec33cc23c5d2cf968aa96e7b516a2c5f1045148470593R123

src/vault.py
saltiyazan: I followed the style of the file.
We could do some refactoring in a different pr.

src/vault.py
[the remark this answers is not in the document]
DanielArndt: > I'm hesitant about this

Completely reasonably. If you think that's the case then lets ignore this for now. It was just something that came to mind, but I don't think I've been working with charms enough to understand the implications.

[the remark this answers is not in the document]
DanielArndt: This might even be a debug log. If you're concerned about noisyness, that's probably the best approach. Either way I'll leave the final call to you, I don't have a strong opinion here at all.

gatici (approved):

PR #70, opened 2023-11-05
feat: Simplifies event management

gruyaume:
# Description

Simplifies event management by using a central `_configure` event handler. Using this handler makes it so that on any state-affecting event (config changed, peer relation changed), we look at the state of the world (relations, configs) and the state of the charm and make the appropriate changes. Through the use of the central event handler, this change gets rid of most `defer()` in the code as well as the need for "wait_for_xyz" methods .

## Notes
1. This PR contains no behaviour change (i.e. integration tests are identical). It's really about "how" events are handled rather than "what" the charm does.
2. Follow up PR's will get rid of defers in the KV handler. We are not doing this here because this PR was large enough as is.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -782,7 +782,7 @@ class VaultCharm(CharmBase):
                 send_ca_cert.remove_certificate(relation.id)
 
     def _ca_certificate_set_in_peer_relation(self) -> bool:
-        """Returns whether CA certificate is stored in Juju secrets."""
+        """Returns whether CA certificate is stored in peer relation data."""
         try:
             ca_private_key, ca_certificate = self._get_ca_certificate_secret_in_peer_relation()
             if ca_private_key and ca_certificate:
--- a/src/charm.py
+++ b/src/charm.py
@@ -792,7 +792,7 @@ class VaultCharm(CharmBase):
         return False
 
     def _initialization_secret_set_in_peer_relation(self) -> bool:
-        """Returns whether initialization secret is stored in Juju secrets."""
+        """Returns whether initialization secret is stored in peer relation data."""
         try:
             root_token, unseal_keys = self._get_initialization_secret_from_peer_relation()
             if root_token and unseal_keys:
--- a/src/charm.py
+++ b/src/charm.py
@@ -39,7 +39,6 @@ from ops.charm import (
 from ops.main import main
 from ops.model import (
     ActiveStatus,
-    MaintenanceStatus,
     ModelError,
     Relation,
     Secret,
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -884,18 +781,25 @@ def _send_ca_cert(self, *, rel_id=None) -> None:
             for relation in self.model.relations.get(SEND_CA_CERT_RELATION_NAME, []):
                 send_ca_cert.remove_certificate(relation.id)
 
-    @property
-    def _vault_ca_certificate_is_stored(self) -> bool:
-        """Returns whether CA certificate is stored in Juju secrets.
+    def _ca_certificate_set_in_peer_relation(self) -> bool:
+        """Returns whether CA certificate is stored in Juju secrets."""
```
gatici, line 785:
```suggestion
        """Returns whether CA certificate is stored in peer relation data."""
```

tests/unit/test_charm.py
```diff
@@ -137,120 +141,19 @@ def _set_other_node_api_address_in_peer_relation(self, relation_id: int, unit_na
             key_values=key_values,
         )
 
-    @patch("charm.generate_vault_unit_certificate")
-    @patch("ops.model.Model.get_binding")
     @patch("ops.model.Container.remove_path")
-    def test_given_not_leader_when_install_then_existing_data_is_removed(
```
gatici, line 143: Is there any test which validates the behavior of removed test "not_leader_when_install" ?

gatici (changes requested):

tests/unit/test_charm.py
gruyaume: On install we don't care anymore whether the unit is leader or not so there's no need to test this specifically. Here's the on install handler:

```python
def _on_install(self, event: InstallEvent):
        """Handler triggered when the charm is installed.

        Sets pebble plan, initializes vault, enable audit device, and unseals vault.
        """
            self.unit.status = WaitingStatus("Waiting to be able to connect to vault unit")
            event.defer()
            return
        self._delete_vault_data()
```

tests/unit/test_charm.py
```diff
@@ -137,120 +141,19 @@ def _set_other_node_api_address_in_peer_relation(self, relation_id: int, unit_na
...
 
         self.harness.charm.on.install.emit()
 
         patch_remove_path.assert_has_calls(
             calls=[call(path="/vault/raft/vault.db"), call(path="/vault/raft/raft/raft.db")]
         )
 
-    @patch("ops.model.Model.get_binding")
-    @patch("charm.generate_vault_unit_certificate")
-    def test_given_non_leader_and_ca_cert_is_created_when_install_then_unit_cert_is_generated(
```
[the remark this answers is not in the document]
gruyaume: The `on_install` hook completely changed, it does not take care of unit certificate generation anymore hence why we removed this test. We have a test `test_given_unit_certificate_not_stored_when_configure_then_unit_certificate_is_generated` that will take care of validating unit certificate generation.

gatici (approved):
LGTM

PR #77, opened 2023-11-22
feat: Add loki log forwarding

Kayra1:
# Description
**What?**

This change adds automatic log forwarding to Vault when integrated with loki.

**Why?**

This will allow users to see their logs on loki instead of having to find them from kubernetes or from the container itself.

**Where?**

This change can be used by all deployments of the vault-k8s charm.

**How?**

On receiving a loki-push-api integration, a pebble layer including `log-forwarding` with the given URL is created and placed in the plan, and the vault container is restarted to include it. On relation departure, this plan is removed and the container restarted again.

**When?**

This change will be released on the next version of the charm, revision 51.

# Checklist:


Kayra1:
I used to use this format for my PR's, I wanted to test it out to see what you guys think also.

lib/charms/loki_k8s/v0/loki_push_api.py
gruyaume, line 1: Let's use `v1` of the lib. It comes with the LogForwarder class making it so that we don't even have to manually interface with pebble's log-targets. 

See this PR for example:
- https://github.com/canonical/sdcore-amf-k8s-operator/pull/96/

gruyaume (changes requested):

lib/charms/loki_k8s/v0/loki_push_api.py
Kayra1: Nice. Done.

PR #82, opened 2023-11-27
feat: Implements backup action

saltiyazan:
# Description

- Implements S3 interface
- Adds backup functions
- Implements backup action
- Adds unit tests

# Checklist:


src/vault.py
```diff
@@ -126,6 +126,16 @@ def enable_audit_device(self, device_type: str, path: str) -> None:
         )
         logger.info("Enabled audit device %s", device_type)
 
+    def create_snapshot(self) -> requests.Response:
+        """Create a snapshot of the Vault data."""
+        logger.info("Created snapshot of the raft backend")
```
gruyaume, line 131: Please log after the action was completed, here it is possible for the `created snapshot ...` log to display and the snapshot not to have been done. This applies in other places too

gruyaume (changes requested):

src/charm.py
```diff
@@ -738,10 +841,107 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            logger.info("Bucket %s exists.", bucket_name)
+            return
+        except ClientError:
+            logger.warning("Bucket %s doesn't exist, creating it.", bucket_name)
+            pass
+        except BotoCoreError as e:
+            logger.warning("Failed to check wether bucket exists. %s", e)
+            raise e
+        try:
+            if region == "us-east-1":
```
gruyaume, line 925: Why hardcode this region?

gruyaume (changes requested):

src/charm.py
```diff
@@ -738,10 +841,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            return session
+        except KeyError as e:
+            logger.warning("Missing required S3 parameter: %s", e)
+            return None
+        except ValueError as e:
+            logger.warning("Error creating resource: %s", e)
+            return None
+
+    def _retrieve_s3_parameters(self) -> Tuple[Dict, List[str]]:
+        """Retrieve S3 parameters from the S3 integrator relation."""
```
gruyaume, line 869: Please give a bit more information in terms of the return

src/charm.py
```diff
@@ -738,10 +841,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            session.resource("s3", endpoint_url=s3_parameters["endpoint"])
+            return session
+        except KeyError as e:
+            logger.warning("Missing required S3 parameter: %s", e)
+            return None
+        except ValueError as e:
+            logger.warning("Error creating resource: %s", e)
+            return None
+
+    def _retrieve_s3_parameters(self) -> Tuple[Dict, List[str]]:
```
gruyaume, line 868: It's not clear to me why we return a tuple, I think I'd rather have 2 methods with independent purposes (`_get_missing_s3_parameters` and `_get_s3_parameters`) instead of having 1 method that does 2 things.

src/charm.py
```diff
@@ -738,10 +841,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+        """
+        s3 = session.resource("s3", endpoint_url=endpoint)
+        try:
+            # Checking if bucket already exists
+            bucket = s3.Bucket(bucket_name)
+            bucket.meta.client.head_bucket(Bucket=bucket_name)
+            logger.info("Bucket %s exists.", bucket_name)
+            return
+        except ClientError:
+            logger.warning("Bucket %s doesn't exist, creating it.", bucket_name)
```
gruyaume, line 919: Why have a warning here? We should have warning in expected contexts, not in normal ones.

src/charm.py
gruyaume: I also don't understand why this `if` is in the try/except, it seems like it should be out of it.

src/charm.py
```diff
@@ -801,6 +999,32 @@ def _initialization_secret_set_in_peer_relation(self) -> bool:
             return False
         return False
 
+    def _create_raft_snapshot(self) -> Optional[bytes]:
+        """Creates a snapshot of Vault.
+
+        Returns:
+            bytes: The snapshot content
```
gruyaume, line 1006: Ain't that a lot? Shouldn't we simply create the snapshot and output the file directory?

gruyaume (changes requested):

src/vault.py
saltiyazan: Logging improved.

src/charm.py
saltiyazan: logging improved.

src/charm.py
saltiyazan: This not documented clearly but here are my findings on this topic:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucketConfiguration.html

Here `us-east-1` is specified to be the default, and it is not listed among the valid values of `LocationConstraint`
I read in some posts that specifying `LocationConstraint` to `us-east-1` can cause an error.

However I tested it and it worked if using S3 compatible storage like `minio` it will work, but in `AWS` we'd get `botocore.exceptions.ClientError: An error occurred (InvalidLocationConstraint) when calling the CreateBucket operation: The specified location-constraint is not valid`

So I think it safer to have this guard.

What do you mean that it needs to be out of the try? the try is around the create function.

Improved to using a constant.

src/charm.py
saltiyazan: Makes sense.
✅

src/charm.py
saltiyazan: Where would that file be saved?

The idea of this function is to call the `create_snapshot` function of Vault's client, but before doing that we are making sure that vault is initialised and unsealed.

Then we get the content of the snapshot in bytes as returned by vault's client.

Then the caller of this function can save the content to a file if that's intended, or save it into an object on S3 storage as done in this case in the action handler

```
...
         snapshot := self._create_raft_snapshot()
...
         self._upload_byte_content_to_s3(
                ....
                content=snapshot,
                ....
          )
....
```

I improved readability a bit now

src/charm.py
gruyaume: Ok, please add some comments or docstrings in the method to explain this because it's not obvious right now and will likely not be for future people working on this method.

src/charm.py
DanielArndt: I think the best approach here would be to return the file-like object (`response.raw`). boto3 should be able to take it as the Body for `put_object()`. Then, the whole file will not be loaded into memory.

src/charm.py
```diff
@@ -431,6 +441,103 @@ def _on_new_vault_kv_client_attached(self, event: NewVaultKvClientAttachedEvent)
...
+        try:
+            self._create_s3_bucket(
+                session=session,
+                region=s3_parameters["region"],
+                bucket_name=s3_parameters["bucket"],
+                endpoint=s3_parameters["endpoint"],
+            )
+            logger.info("Created S3 bucket %s", s3_parameters["bucket"])
+        except KeyError as e:
+            logger.warning("Missing required S3 parameter: %s", e)
```
ghislainbourgeois, line 480: I think we should log at the `error` level here, we are failing the action right after.

src/charm.py
```diff
@@ -431,6 +441,103 @@ def _on_new_vault_kv_client_attached(self, event: NewVaultKvClientAttachedEvent)
...
+            event.fail(message="Failed to create S3 bucket.")
+            return
+        except (BotoCoreError, ClientError) as e:
+            logger.warning("Failed to create S3 bucket: %s", e)
```
ghislainbourgeois, line 484: I think we should log at the `error` level here, we are failing the action right after.

src/charm.py
```diff
@@ -738,10 +845,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+        try:
+            session = boto3.session.Session(
+                aws_access_key_id=s3_parameters["access-key"],
+                aws_secret_access_key=s3_parameters["secret-key"],
+                region_name=s3_parameters["region"],
+            )
+            session.resource("s3", endpoint_url=s3_parameters["endpoint"])
+            return session
+        except KeyError as e:
+            logger.warning("Missing required S3 parameter: %s", e)
```
ghislainbourgeois, line 866: I think we should log at the `error` level here, we are failing the action right after.

src/charm.py
```diff
@@ -738,10 +845,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            return None
+        except ValueError as e:
+            logger.warning("Error creating resource: %s", e)
```
ghislainbourgeois, line 869: I think we should log at the `error` level here, we are failing the action right after.

src/charm.py
```diff
@@ -738,10 +845,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            # Checking if bucket already exists
+            bucket = s3.Bucket(bucket_name)
+            bucket.meta.client.head_bucket(Bucket=bucket_name)
+            logger.info("Bucket %s exists.", bucket_name)
+            return
+        except ClientError:
+            logger.info("Bucket %s doesn't exist, creating it.", bucket_name)
+            pass
+        except BotoCoreError as e:
+            logger.warning("Failed to check wether bucket exists. %s", e)
```
ghislainbourgeois, line 927: I think we should log at the `error` level here, we are re-raising right after, and this will fail the action.

src/charm.py
```diff
@@ -738,10 +845,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            raise e
+        try:
+            if region == DEFAULT_REGION:
+                bucket = bucket.create()
+            else:
+                bucket.create(CreateBucketConfiguration={"LocationConstraint": region})
+            bucket.wait_until_exists()
+        except (BotoCoreError, ClientError) as error:
+            logger.warning("Couldn't create bucket named '%s' in region=%s.", bucket_name, region)
```
ghislainbourgeois, line 936: I think we should log at the `error` level here, we are failing the action right after.

ghislainbourgeois (changes requested):

src/charm.py
```diff
@@ -6,15 +6,22 @@
...
 """
+import datetime
 import json
 import logging
 import socket
 from typing import Dict, List, Optional, Tuple
 
+import boto3  # type: ignore[import-untyped]
 import hcl  # type: ignore[import-untyped]
+from botocore.exceptions import (  # type: ignore[import-untyped]
```
DanielArndt, line 17: Same thing here, `types-botocore`.

src/charm.py
saltiyazan: I see the point @gruyaume 
I fixed it to use the approach proposed by @DanielArndt

src/charm.py
```diff
@@ -801,6 +1003,32 @@ def _initialization_secret_set_in_peer_relation(self) -> bool:
...
+
+        Returns:
+            bytes: The snapshot content
+        """
+        vault = Vault(url=self._api_address, ca_cert_path=self._get_ca_cert_location_in_charm())
+        if not vault.is_initialized():
+            logger.warning("Vault is not initialized, cannot create snapshot")
+            return None
+        if not vault.is_api_available():
+            logger.warning("Vault API is not available, cannot create snapshot")
```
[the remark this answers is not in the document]
saltiyazan: ✅

src/charm.py
```diff
@@ -801,6 +1003,32 @@ def _initialization_secret_set_in_peer_relation(self) -> bool:
...
 
+    def _create_raft_snapshot(self) -> Optional[bytes]:
+        """Creates a snapshot of Vault.
+
+        Returns:
+            bytes: The snapshot content
+        """
+        vault = Vault(url=self._api_address, ca_cert_path=self._get_ca_cert_location_in_charm())
+        if not vault.is_initialized():
+            logger.warning("Vault is not initialized, cannot create snapshot")
```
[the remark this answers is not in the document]
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

src/charm.py
```diff
@@ -738,10 +845,105 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
...
+            return
+        except ClientError:
+            logger.info("Bucket %s doesn't exist, creating it.", bucket_name)
+            pass
+        except BotoCoreError as e:
+            logger.warning("Failed to check wether bucket exists. %s", e)
+            raise e
+        try:
+            if region == DEFAULT_REGION:
+                bucket = bucket.create()
```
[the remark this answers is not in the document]
saltiyazan: Added.

src/charm.py
saltiyazan: ✅

src/charm.py
```diff
@@ -6,15 +6,22 @@
 
 For more information on Vault, please visit https://www.vaultproject.io/.
 """
+import datetime
 import json
 import logging
 import socket
 from typing import Dict, List, Optional, Tuple
 
+import boto3  # type: ignore[import-untyped]
```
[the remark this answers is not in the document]
saltiyazan: ✅

src/charm.py
```diff
@@ -431,6 +439,103 @@ def _on_new_vault_kv_client_attached(self, event: NewVaultKvClientAttachedEvent)
...
+            self._create_s3_bucket(
+                session=session,
+                region=s3_parameters["region"],
+                bucket_name=s3_parameters["bucket"],
+                endpoint=s3_parameters["endpoint"],
+            )
+            logger.info("Created S3 bucket %s", s3_parameters["bucket"])
+        except KeyError as e:
+            logger.error("Missing required S3 parameter: %s", e)
+            event.fail(message="Failed to create S3 bucket.")
```
gruyaume, line 479: Haven't we already looked for those keys earlier at `missing_parameters = self._get_missing_s3_parameters()`?

src/charm.py
```diff
@@ -738,10 +843,108 @@ def _get_initialization_secret_from_peer_relation(self) -> Tuple[str, List[str]]
         except (TypeError, SecretNotFoundError, AttributeError):
             raise PeerSecretError(secret_name=VAULT_INITIALIZATION_SECRET_LABEL)
 
+    def _create_s3_session(self, s3_parameters: dict) -> Optional[boto3.session.Session]:
```
gruyaume, line 846: It feels like all the `S3` related actions could be in a class of their own

```
class S3():
    def __init__(self, access_key: str, secret_key: str, region: str):
        self.session = self._create_session()
    
    def _create_session():
        ...

    def create_bucket():
        ...

    def upload_content():
        ...

```

This should simplify the code and tests

gruyaume, line 846: I suspect the answer is yes since we're using it but `boto3` should work for non-aws S3?

gruyaume (changes requested):

tests/unit/test_charm.py
```diff
@@ -987,3 +994,351 @@ def test_given_ca_cert_is_not_stored_when_certificate_transfer_relation_joins_th
...
+    ):
+        patch_is_initialized.return_value = True
+        patch_is_api_available.return_value = True
+        self.harness.set_can_connect(container=self.container_name, val=True)
+        self.harness.add_storage(storage_name="certs", attach=True)
+        self.harness.charm._create_raft_snapshot()
+        patch_create_snapshot.assert_not_called()
+
+    def test_given_cannot_connect_to_session_when_create_s3_bucket_then_exception_is_raised(self):
+        with self.assertRaises(Exception):
```
DanielArndt, line 1099: Can we be more specific about what type of exception we expect here? This can bite us if the method starts failing for a different reason than we expect it to, as the test  will continue to pass.

tests/unit/test_charm.py
```diff
@@ -987,3 +994,351 @@ def test_given_ca_cert_is_not_stored_when_certificate_transfer_relation_joins_th
...
+                bucket_name="whatever bucket",
+                region="whatever region",
+                endpoint="whatever endpoint",
+            )
+
+    def test_given_bucket_already_exists_when_create_s3_bucket_then_bucket_not_created(
+        self,
+    ):
+        patch_session = Mock()
+        patch_session.resource.Bucket.meta.client.head_bucket.return_value = Mock()
```
DanielArndt, line 1111: You shouldn't need this line, because a Mock returns a Mock for each property that is accessed.
```suggestion
```

src/charm.py
```diff
@@ -431,6 +439,103 @@ def _on_new_vault_kv_client_attached(self, event: NewVaultKvClientAttachedEvent)
...
+        if not backup_key:
+            event.fail(message="Failed to upload backup to S3 bucket.")
+            return
+        logger.info("Backup uploaded to S3 bucket %s", s3_parameters["bucket"])
+        event.set_results({"backup-id": backup_key})
+
+    def _upload_content_to_s3(
+        self,
+        session: boto3.session.Session,
+        content: requests.Response,
```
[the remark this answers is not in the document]
ghislainbourgeois: Probably something here could be used: https://docs.python.org/3/library/typing.html#typing.IO

src/charm.py
```diff
@@ -431,6 +439,103 @@ def _on_new_vault_kv_client_attached(self, event: NewVaultKvClientAttachedEvent)
...
+        Returns:
+            str: The S3 key of the uploaded content.
+        """
+        key = self._get_backup_key()
+        try:
+            s3 = session.resource("s3", endpoint_url=endpoint)
+            bucket = s3.Bucket(bucket_name)
+            bucket.put_object(Key=key, Body=content)
+        except (BotoCoreError, ClientError) as e:
+            logger.warning("Error uploading content to bucket %s: %s", bucket_name, e)
```
ghislainbourgeois, line 525: Found another one that might be better at the `error` level.

ghislainbourgeois (changes requested):
Small comment

src/charm.py
saltiyazan: Yes it works, tested on `Minio`.

tests/unit/test_charm.py
saltiyazan: removed

src/charm.py
saltiyazan: ✅

tests/unit/test_charm.py
saltiyazan: Tests improved and moved to `test_s3_helpers.py`

src/charm.py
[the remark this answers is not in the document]
saltiyazan: ✅

src/charm.py
```diff
@@ -801,6 +1004,32 @@ def _initialization_secret_set_in_peer_relation(self) -> bool:
             return False
         return False
 
+    def _create_raft_snapshot(self) -> Optional[requests.Response]:
+        """Creates a snapshot of Vault.
+
+        Returns:
+            bytes: The snapshot content
```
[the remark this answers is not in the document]
saltiyazan: fixed

src/charm.py
```diff
@@ -801,6 +1004,32 @@ def _initialization_secret_set_in_peer_relation(self) -> bool:
             return False
         return False
 
+    def _create_raft_snapshot(self) -> Optional[requests.Response]:
```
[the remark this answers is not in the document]
saltiyazan: ✅

test-requirements.txt
```diff
@@ -17,3 +17,5 @@ types-PyYAML
 types-setuptools
 types-toml
 types-requests
+types-boto3
+types-boto3
```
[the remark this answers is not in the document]
saltiyazan: ✅

src/charm.py
saltiyazan: Correct, fixed.

src/s3_helpers.py
gruyaume, line 1: Why not have everything here in a class instead of individual functions?

src/charm.py
```diff
@@ -65,6 +70,9 @@
 CA_CERTIFICATE_JUJU_SECRET_LABEL = "vault-ca-certificate"
 SEND_CA_CERT_RELATION_NAME = "send-ca-cert"
 VAULT_INITIALIZATION_SECRET_LABEL = "vault-initialization"
+S3_RELATION_NAME = "s3-parameters"
+DEFAULT_REGION = "us-east-1"  # It is used as default in boto3
```
gruyaume, line 74: We have this set here as well as in the s3 file, we should probably only keep one to avoid having to guess which one is the right one.

tests/unit/test_charm.py
```diff
@@ -987,3 +994,302 @@ def test_given_ca_cert_is_not_stored_when_certificate_transfer_relation_joins_th
...
+        self.harness.charm._on_create_backup_action(event)
+        event.fail.assert_called_with(message="Failed to create raft snapshot.")
+
+    @patch("vault.Vault.is_sealed", new=Mock)
+    @patch("vault.Vault.unseal", new=Mock)
+    @patch("vault.Vault.create_snapshot", new=Mock)
+    @patch("vault.Vault.is_initialized")
+    @patch("vault.Vault.is_api_available")
+    @patch(f"{S3_LIB_PATH}.S3Requirer.get_s3_connection_info")
+    @patch("boto3.session.Session")
```
gruyaume, line 1212: We shouldn't have to mock this anymore since we are not using it directly in the charm code. This applies in other tests as well

gruyaume (changes requested):

src/charm.py
saltiyazan: Now it's only handled in the s3 file.

tests/unit/test_charm.py
saltiyazan: Keeping the mock of this external dependency would make the tests faster.
However I removed it and I am following the same pattern used with mocking vault client.

src/s3_session.py
```diff
@@ -0,0 +1,117 @@
...
+                aws_access_key_id=self.access_key,
+                aws_secret_access_key=self.secret_key,
+                region_name=self.region,
+            )
+            self.session.resource("s3", endpoint_url=self.endpoint)
+        except (ClientError, BotoCoreError, ValueError) as e:
+            logger.error("Error creating AWS session: %s", e)
+            raise e
+
+    def create_s3_bucket(self, bucket_name: str) -> bool:
```
gruyaume, line 43:
```suggestion
    def create_bucket(self, bucket_name: str) -> bool:
```

gruyaume (changes requested):
No need for `s3` in the method names since they're all under an `S3` class.

tests/unit/test_vault.py
```diff
@@ -141,6 +141,25 @@ def test_given_audit_device_is_not_yet_enabled_when_enable_audit_device_then_dev
...
         )
 
+    @patch("hvac.api.system_backend.Raft.take_raft_snapshot")
+    def test_given_raft_backend_when_create_snapshot_then_take_raft_snapshot_is_called(
+        self,
+        patch_take_raft_snapshot,
+    ):
+        vault = Vault(url="http://whatever-url", ca_cert_path="whatever path")
+        vault.create_snapshot()
+        patch_take_raft_snapshot.assert_called_once()
```
gruyaume, line 151: I'm not sure there's much value in tests like these

tests/unit/test_charm.py
```diff
@@ -987,3 +998,271 @@ def test_given_ca_cert_is_not_stored_when_certificate_transfer_relation_joins_th
...
+        patch_is_initialized.return_value = True
+        self.harness.set_can_connect(container=self.container_name, val=True)
+        self.harness.add_storage(storage_name="certs", attach=True)
+        peer_relation_id = self._set_peer_relation()
+        self._set_initialization_secret_in_peer_relation(
+            relation_id=peer_relation_id,
+            root_token="root token content",
+            unseal_keys=["unseal_keys"],
+        )
+        self.harness.charm._create_raft_snapshot()
```
gruyaume, line 1048: We shouldn't test our private methods directly, here the public interface is really the Juju action handler (i.e. `_on_create_backup_action`) and this is what should be tested. This applies to all tests aimaing at `_create_raft_snapshot` directly.

gruyaume (changes requested):

tests/unit/test_vault.py
saltiyazan: Removed

tests/unit/test_charm.py
saltiyazan: Updated

ghislainbourgeois (approved):

tests/unit/test_s3_session.py
```diff
@@ -0,0 +1,169 @@
...
+            region=valid_s3_parameters["region"],
+            endpoint=valid_s3_parameters["endpoint"],
+        )
+
+        s3.create_bucket(bucket_name="whatever-bucket")
+
+        mock_bucket.create.assert_called_once()
+
+    @patch("boto3.session.Session")
+    def test_bucket_does_not_exist_when_upload_content_then_content_not_uploaded(
```
gruyaume, line 111:
```suggestion
    def test_given_bucket_does_not_exist_when_upload_content_then_content_not_uploaded(
```

tests/unit/test_s3_session.py
```diff
@@ -0,0 +1,169 @@
...
+        }
+        s3 = S3(
+            access_key=valid_s3_parameters["access-key"],
+            secret_key=valid_s3_parameters["secret-key"],
+            region=valid_s3_parameters["region"],
+            endpoint=valid_s3_parameters["endpoint"],
+        )
+        self.assertIsInstance(s3.session, boto3.session.Session)
+
+    def test_given_invalid_endpoint_when_create_s3_session_then_session_not_created(self):
```
gruyaume, line 31:
```suggestion
    def test_given_invalid_endpoint_when_create_s3_session_then_valueerror_is_raised(self):
```

gruyaume (approved):

PR #93, opened 2023-12-04
fix: Update README.md

Kayra1:
I was looking over [this](https://developer.hashicorp.com/vault/docs/commands#vault_cacert) last week, and noticed that CAPATH if for directories and CACERT is for files, even though CAPATH also works for single CA files. We should probably use the one for single files if the tutorial wants us to use a single file.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,10 +148,10 @@ ck0i0h3q457c7bgte4kg:
     [...]
 ```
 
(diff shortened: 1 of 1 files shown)


gruyaume:
This PR likely fails because of the asyncio issue. This is addressed in PR #94

ghislainbourgeois (approved):

PR #94, opened 2023-12-04
chore: Freezes dependencies

gruyaume:
# Description

Having our dependencies not freezed results in the team not knowing when one of them changes and often combining a feature change with a dependency change without realizing it. Sometimes this results in unexpected CI failures and having to find and fix the dependency related problem.

Here we use `pip freeze -r requirements.txt` to freeze python dependencies. This permits individual CI runs for dependency update through dependabot and avoids having our CI fail for unpredictable reasons.

# Checklist:


requirements.txt
ghislainbourgeois, line 1: What do you think about using `pip-compile` instead of `pip freeze`. It uses a `requirements.in` file that we maintain with our direct dependencies, and then solves the dependencies and pins them in a `requirements.txt` file. It would keep our direct dependencies clearer IMO.

Also, `pip freeze` when not used properly can add dependencies just because they are installed in the current environment.

This is also one of the recommended tools in [TE030](https://docs.google.com/document/d/17CpHK9bJcLBo8T7b8hgt9j0WMG6Tbgth0p7sycXAs6I).

https://pip-tools.readthedocs.io/en/latest/

gruyaume: I don't have a strong opinion here, I'll make the switch to `pip-compile`

gruyaume: @ghislainbourgeois In our case however we have 2 requirements files:
- requirements.txt
- test-requirements.txt

This means that we'd also need 2 new `.in` files. However I don't see how we can have `pip compile` use a different file from `requirements.in` (`test-requirements.in` for instance).

gruyaume: Ok now that I tried it, I see those 2 cons with using `pip-compile` instead of `pip freeze`:
- We add a new tool that we all need to understand. When we add dependencies to our projects, we need to add them in a new file (requirements.in), install `pip-tools` and run `pip compile`
- We don't maintain our dependencies' dependencies. For example, this solution would not have prevented the asyncio dependency issue we have now (reference here: https://github.com/canonical/sdcore-upf-operator/pull/61/files)

ghislainbourgeois: I do not have a super strong opinion either way. I have a slight preference for `pip-compile` because it is more explicit.

Maybe this warrants a quick team discussion at the daily to align?

PR #107, opened 2023-12-12
feat: Implements list and restore actions

saltiyazan:
# Description

- Implements list-backups action
- Implements restore-backup action
- Allows the initialization secret to be updated
- Adds unit test cases
- Improves unit testing

# Checklist:


src/charm.py
```diff
@@ -498,14 +502,195 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:
...
+        vault.set_token(token=current_root_token)
+        if vault.is_sealed():
+            vault.unseal(unseal_keys=current_unseal_keys)
+
+        self._set_initialization_secret_in_peer_relation(
+            root_token=restore_root_token, unseal_keys=restore_unseal_keys
+        )
+        try:
+            # hvac vault client expects bytes or a file-like object to restore the snapshot
+            # StreamingBody is implements the read() method
```
ghislainbourgeois, line 665:
```suggestion
            # StreamingBody implements the read() method
```

src/charm.py
```diff
@@ -498,14 +502,195 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:
...
+            s3 = S3(
+                access_key=s3_parameters["access-key"],
+                secret_key=s3_parameters["secret-key"],
+                endpoint=s3_parameters["endpoint"],
+                region=s3_parameters.get("region"),
+            )
+        except (BotoCoreError, ClientError, ValueError) as e:
+            logger.error("Failed to create S3 session: %s", e)
+            event.fail(message="Failed to create S3 session.")
+            return
```
ghislainbourgeois, line 540: This exact code is repeated 3 times now, once for each action. Would it make sense to extract it to a separate method? It could make it simpler to maintain in the future, for things like changing log messages or renaming parameters. WDYT?

ghislainbourgeois (changes requested):
Overall looks good, I have some small comments.

src/charm.py
saltiyazan: I'm hesitant between the following approaches.
1. Having a function that we pass the event to as a parameter and it does the requirement check
My problem with this approach is that this function will be coupled to the context of the action handlers.
```
def _check_backup_restore_actions_s3_requirements(self, event: ActionEvent) -> None:
        """Checks the conditions for creating S3 in backup and restore actions.

        Args:
            event: ActionEvent

        Returns:
            bool: True if the conditions are met, False otherwise.
        """
        if not self._is_relation_created(S3_RELATION_NAME):
            logger.error("S3 relation not created. Failed to run backup/restore operation.")
            event.fail(message="S3 relation not created. Failed to run backup/restore operation.")
            return

        if not self.unit.is_leader():
            logger.error("Only leader unit can run backup/restore operation.")
            event.fail(message="Only leader unit can run backup/restore operation.")
            return

        missing_parameters = self._get_missing_s3_parameters()
        if missing_parameters:
            logger.error("S3 parameters missing. %s", missing_parameters)
            event.fail(message=f"S3 parameters missing. {missing_parameters}")
```

2. Having a more specific requirements check that will not validate everything, but will be scoped to the requirements of S3 (similar one for Vault requirements)
This is better as it's not coupled and can be used in different contexts if needed, the problem with is that it will need to return a message and some of the checks will be kept in the action handler causing a few cases of code duplication (for example checking leader)

```
def _check_s3_requirements(self) -> Tuple[bool, Optional[str]]:
        """Validates the conditions for creating S3.

        It will check if the S3 relation is created
            and if the required S3 parameters are set.

        Returns:
            bool: True if the conditions are met, False otherwise.
        """
        if not self._is_relation_created(S3_RELATION_NAME):
            return False, "S3 relation not created."

        missing_parameters = self._get_missing_s3_parameters()
        if missing_parameters:
            return False, f"S3 parameters missing. {missing_parameters}"

        return True, None
```

I feel there is a trade-off here, and personally I prefer not to couple functions to each other, therefore, I prefer the second choice and I have pushed a commit using this solution (wiht the Vault requirements checks too)

Let me know what do you think and if you see a different approach or if we can do something else to improve this.

ghislainbourgeois (approved):

src/charm.py
ghislainbourgeois: This works for me.

PR #118, opened 2024-01-16
chore: Improves logging in vault.py

saltiyazan:
# Description

Adds logging to functions that call `read_health_status`

# Checklist:

### DIFF
--- a/src/vault.py
+++ b/src/vault.py
@@ -17,7 +17,7 @@ RAFT_STATE_ENDPOINT = "v1/sys/storage/raft/autopilot/state"
 
 
 class VaultClientError(Exception):
-    """Custom error."""
+    """Custom exception to pass errors with Vault client calls."""
 
     def __init__(self, message: str = "An error occurred while interacting with Vault client."):
         self.message = message
@@ -45,7 +45,7 @@ class Vault:
             )
             logger.info("Vault is initialized")
             return initialize_response["root_token"], initialize_response["keys"]
-        except VaultError as e:
+        except (RequestException, VaultError) as e:
             logger.error("Error initializing Vault: %s", e)
             return None
 
@@ -81,7 +81,7 @@ class Vault:
         try:
             self._client.sys.read_health_status()
             return True
-        except RequestException as e:
+        except (RequestException, VaultError) as e:
             logger.error("Vault API is not available: %s", e)
             return False
 
--- a/src/vault.py
+++ b/src/vault.py
@@ -5,19 +5,23 @@
 """Contains all the specificities to communicate with Vault through its API."""
 
 import logging
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 import hvac  # type: ignore[import-untyped]
 import requests
+from hvac.exceptions import VaultError  # type: ignore[import-untyped]
+from requests.exceptions import RequestException
 
 logger = logging.getLogger(__name__)
 RAFT_STATE_ENDPOINT = "v1/sys/storage/raft/autopilot/state"
 
 
-class VaultError(Exception):
-    """Exception raised for Vault errors."""
+class VaultClientError(Exception):
+    """Custom error."""
 
-    pass
+    def __init__(self, message: str = "An error occurred while interacting with Vault client."):
+        self.message = message
+        super().__init__(self.message)
 
 
 class Vault:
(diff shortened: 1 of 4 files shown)


src/vault.py
```diff
@@ -28,17 +32,22 @@ def __init__(self, url: str, ca_cert_path: str):
...
-        return initialize_response["root_token"], initialize_response["keys"]
+        try:
+            initialize_response = self._client.sys.initialize(
+                secret_shares=secret_shares, secret_threshold=secret_threshold
+            )
+            logger.info("Vault is initialized")
+            return initialize_response["root_token"], initialize_response["keys"]
+        except (RequestException, VaultError) as e:
+            logger.error("Error initializing Vault: %s", e)
+            return None
```
gruyaume, line 50: Why do we return None in this case but in other exception cases we raise `VaultClientError`?

DanielArndt:
I haven't gone through this in full detail, but I don't think this is the best approach. It's very verbose, not just in code, but also we're logging the same error multiple times in different places.

I don't know if you tried this, but I think the best way to solve this would be to reproduce some of the errors we were experiencing, and figuring out why the underlying error was never reported. This might be as simple as adding a log in the `try..except` blocks where we are effectively suppressing the errors (or we're re-branding them or something). If the error bubbles its way up to the top, I would expect we would get the full log.

I'll give this a closer look tomorrow, maybe I'm off base.

src/vault.py
saltiyazan: We could do that too, but in most cases where the exception was raised again we can't do much with the return value, on the other hand in this case the return value has a clear meaning. I can change it for consistency if that's preferred.
We could do that too, but in most of the other functions we can't use the return value for this purpose, some don't return anything, while others return bool values where both True and False have different meanings than "error". but in this case the return value clearly carries the information. If you prefer to go towards consistency among all functions I can change it.

saltiyazan:
> I haven't gone through this in full detail, but I don't think this is the best approach. It's very verbose, not just in code, but also we're logging the same error multiple times in different places.
> 
> I don't know if you tried this, but I think the best way to solve this would be to reproduce some of the errors we were experiencing, and figuring out why the underlying error was never reported. This might be as simple as adding a log in the `try..except` blocks where we are effectively suppressing the errors (or we're re-branding them or something). If the error bubbles its way up to the top, I would expect we would get the full log.
> 
> I'll give this a closer look tomorrow, maybe I'm off base.

I have to agree with you, it is pretty verbose and many lines of code to handle issues that might never happen.

However I don’t think we should be pushing the error handling to the charm code since we have this wrapper class and we can abstract such things here.

I can think of some solutions to extract some parts into a common place like using decorators.

Yes, I have reproduced the issue, it raises when checking api availability with `is_api_available` we catch a `ConnectionError` and return false, but we don’t log the exception, so we just get the fact that the API was not available and we don’t know why.

And that is to some extent an expected behavior, if there is an issue connecting to Vault we can assume that the API is not available.

What we can do here is to keep the same behavior and log the issue, or change the behavior of `is_api_available` and look at the response and return, raise or log accordingly.

For the rest of the functions the issues we can get when interacting with Vault are of the same nature, but some functions don’t handle exceptions and don’t return values, so to reduce verbosity we could add logging to whichever functions that return a value and keep it as is for those that don’t and then if an exception is raised the charm crashes.

What do you think? @DanielArndt 
Do you think we should keep this minimalistic only to functions that already handled exceptions?

DanielArndt:
> However I don’t think we should be pushing the error handling to the charm code since we have this wrapper class and we can abstract such things here.

I agree.
 
> So the solution could be as simple as adding a log message in `is_api_available`, or to look into the response and decide if we should return false or raise.

I think adding a log here is exactly what we need. All the better is we can look at the response and only log that when it's something unexpected. The behaviour of the charm otherwise was fine, but it swallowed the underlying error message so we couldn't see it. We didn't know what was actually wrong. The error message was misleading without seeing the underlying error that it masked. 

We may still want to add some other `try..except` blocks in `vault.py` that prevent errors from bubbling up from the client,.

That being said, we should put some thought into where each log belongs. It's going to be context dependent in some cases where there isn't an error from `vault.py`'s perspective, but it is an error from the charm code's perspective. In this case I think it makes sense to log in the charm instead. 

For example a [fictitious] function like `is_raft_enabled()` might detect if raft is enabled by making some call that raises an exception, but that's not an error from its perspective. If the charm really expects raft to be enabled at the point that function is called, the charm should be the one to log the error.

> However we can get exceptions of a similar nature when the interaction with vault fails in other functions, we could do what I did and handle errors and add logs in all functions, or we can be selective and do it only in functions like `is_api_available` where we already depend on the exception but don't log. The second option would significantly reduce the verbosity, but we'd be risking the charm to crash in some cases. For example we add logs to `is_api_available` and keep `unseal` as originally was, now if there is an issue interacting with vault when we call `is_api_available` we will know what the issue was from the logs, but if something goes wrong later on and we can't do `unseal` at a specific moment the charm will crash instead of us deciding what should the charm do until the next event/hook. I don't think this would be that bad and it would reduce this change to a few lines only, but it's a tradeoff.

In general, I prefer to let unexpected things break rather than to hide them, so if it is truly unexpected I think it _should_ crash (in most cases). But, if its something we could predict such as a network timeout, we should handle those scenarios, log them, and do whatever we can to keep the charm resilient to predictable or possibly intermittent issues.

Ideally, we'd separate unexpected errors, transient errors, and permanent errors and handle each case differently, but I'm not sure we have a great pattern for dealing with transient errors at this time.

This was a lot of rant and I'm not sure how clear I'm being here. Would love to jump on a call with you in the morning and chat about this.

PR #124, opened 2024-01-26
chore: Uses pip-tools and dependabot to manage dependencies and versions

saltiyazan:
# Description

- Removes renovate and replaces it with dependabot to automate dependency version updates
- Replaces requirements.txt files with requirements.in, requirements.txt is generated using pip-tools" 

Files were generated using the following commands:
- `pip-compile requirements.in`
- `pip-compile test-requirements.in -c requirements.txt`

# Checklist:


gruyaume (approved):

DanielArndt:
I'm going to merge this, because it should fix the broken integration tests which are, for some reason, suddenly pulling in a higher version of pydantic which is breaking the build. Rather than pin the versions in a separate PR, I'll take the fallout from this if anything goes wrong :smile:

PR #142, opened 2024-01-31
feat: Manual Unsealing

### LINKED ISSUE #242: Vault raft autopilot in unhealthy state after juju unit removal
#### Describe the bug
Vault goes into unhealthy state after juju units get removed. It looks like this is because the removed nodes do not get properly removed from the raft cluster.

#### To Reproduce

1. Deploy Vault with 3 units
3. Scale to 5 units
4. Check vault autopilot state (`vault operator raft autopilot state`)

#### Expected behavior

Vault remains healthy, removed nodes get removed from raft cluster.

#### Logs

```
Healthy:                         false
Failure Tolerance:               0
Leader:                          sec-vault-k8s/0
Voters:
   sec-vault-k8s/0
   sec-vault-k8s/1
   sec-vault-k8s/2
   sec-vault-k8s/4
   sec-vault-k8s/3
Servers:
   sec-vault-k8s/0
      Name:              sec-vault-k8s/0
      Address:           vault-k8s-0.vault-k8s-endpoints.sec.svc.cluster.local:8201
      Status:            leader
      Node Status:       alive
      Healthy:           true
      Last Contact:      0s
      Last Term:         3
      Last Index:        361
      Version:           1.15.6
      Node Type:         voter
   sec-vault-k8s/1
      Name:              sec-vault-k8s/1
      Address:           vault-k8s-1.vault-k8s-endpoints.sec.svc.cluster.local:8201
      Status:            voter
      Node Status:       alive
      Healthy:           true
      Last Contact:      15.691837ms
      Last Term:         3
      Last Index:        361
      Version:           1.15.6
      Node Type:         voter
   sec-vault-k8s/2
      Name:              sec-vault-k8s/2
      Address:           vault-k8s-2.vault-k8s-endpoints.sec.svc.cluster.local:8201
      Status:            voter
      Node Status:       alive
      Healthy:           true
      Last Contact:      1.425088036s
      Last Term:         3
      Last Index:        361
      Version:           1.15.6
      Node Type:         voter
   sec-vault-k8s/3
      Name:              sec-vault-k8s/3
      Address:           vault-k8s-3.vault-k8s-endpoints.sec.svc.cluster.local:8201
      Status:            voter
      Node Status:       alive
      Healthy:           false
      Last Contact:      1m27.460585464s
      Last Term:         3
      Last Index:        359
      Version:           1.15.6
      Node Type:         voter
   sec-vault-k8s/4
      Name:              sec-vault-k8s/4
      Address:           vault-k8s-4.vault-k8s-endpoints.sec.svc.cluster.local:8201
      Status:            voter
      Node Status:       alive
      Healthy:           false
      Last Contact:      1m26.792905756s
      Last Term:         3
      Last Index:        359
      Version:           1.15.6
      Node Type:         voter
```

While we do see the logs that the db's get removed, we never see that the node gets removed from the raft cluster:
```
unit-vault-k8s-3: 16:08:44 INFO juju.worker.uniter.operation ran "vault-peers-relation-departed" hook (via hook dispatching script: dispatch)
unit-vault-k8s-4: 16:08:44 INFO juju.worker.uniter.operation ran "vault-peers-relation-departed" hook (via hook dispatching script: dispatch)
unit-vault-k8s-3: 16:08:45 INFO juju.worker.uniter.operation ran "vault-pki-relation-broken" hook (via hook dispatching script: dispatch)
unit-vault-k8s-4: 16:08:45 INFO juju.worker.uniter.operation ran "vault-peers-relation-departed" hook (via hook dispatching script: dispatch)
unit-vault-k8s-3: 16:08:46 INFO juju.worker.uniter.operation ran "stop" hook (via hook dispatching script: dispatch)
unit-vault-k8s-4: 16:08:46 INFO juju.worker.uniter.operation ran "stop" hook (via hook dispatching script: dispatch)
unit-vault-k8s-3: 16:08:47 INFO unit.vault-k8s/3.juju-log Vault CA certificate not found
unit-vault-k8s-4: 16:08:47 INFO unit.vault-k8s/4.juju-log Vault CA certificate not found
unit-vault-k8s-4: 16:08:47 INFO unit.vault-k8s/4.juju-log Removed Vault's main database
unit-vault-k8s-4: 16:08:47 INFO unit.vault-k8s/4.juju-log Removed Vault's Raft database
unit-vault-k8s-3: 16:08:47 INFO unit.vault-k8s/3.juju-log Removed Vault's main database
unit-vault-k8s-3: 16:08:47 INFO unit.vault-k8s/3.juju-log Removed Vault's Raft database
unit-vault-k8s-4: 16:08:48 INFO juju.worker.uniter.operation ran "remove" hook (via hook dispatching script: dispatch)
unit-vault-k8s-3: 16:08:48 INFO juju.worker.uniter.operation ran "remove" hook (via hook dispatching script: dispatch)
```
#### Environment

- Charm / library version (if relevant): 1.15/edge (rev. 176)
- Juju version (output from `juju --version`): 3.4.0-genericlinux-amd64
- Cloud Environment: MicroK8s
- Kubernetes version (output from `kubectl version --short`): MicroK8s v1.28.7 revision 6532

Kayra1:
Fixes #242 
# Description

Implements manual unseal as described in the manual unseal spec.

# Checklist:


lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -368,7 +379,7 @@ def restore_snapshot(self, snapshot: bytes) -> requests.Response:
         Uses force_restore_raft_snapshot to restore the snapshot
         even if the unseal key used at backup time is different from the current one.
         """
-        return self._client.sys.force_restore_raft_snapshot(snapshot)
+        return self._client.sys.restore_raft_snapshot(snapshot)
```
gruyaume, line 382: This change doesn't seem related to the scope of the PR (manual unseal)

src/charm.py
```diff
@@ -1086,71 +1066,46 @@ def _create_raft_snapshot(self) -> Optional[IO[bytes]]:
...
-            self._set_initialization_secret(
-                root_token=current_root_token, unseal_keys=current_unseal_keys
-            )
+            logger.error("Failed to restore snapshotes: %s", response.json())
             return False
-        vault.authenticate(Token(restore_root_token))
-        vault.unseal(unseal_keys=restore_unseal_keys)
+
+        # Check if the approle can still login
+        # If not, remove secret
```
gruyaume, line 1102: Why those comments here?

src/charm.py
```diff
@@ -1086,71 +1066,46 @@ def _create_raft_snapshot(self) -> Optional[IO[bytes]]:
...
-            root_token=restore_root_token, unseal_keys=restore_unseal_keys
-        )
+
         try:
             # hvac vault client expects bytes or a file-like object to restore the snapshot
             # StreamingBody implements the read() method
             # so it can be used as a file-like object in this context
             response = vault.restore_snapshot(snapshot)  # type: ignore[arg-type]
         except Exception as e:
             # If restore fails for any reason, we reset the initialization secret
```
gruyaume, line 1094: Do we?

actions.yaml
```diff
@@ -1,6 +1,17 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+authorize-charm:
+  description: >-
+    Authorizes the charm to be able to interact with Vault to manage its operations. The root token
```
gruyaume, line 6: We don't need the `root` token, all we need is a token. It can be created via
```
vault token create -ttl=10m
```

actions.yaml
```diff
@@ -1,6 +1,17 @@
...
+    is required for Vault to use to create its own policy and approle once, then discard.
+  params:
+    root-token:
```
gruyaume, line 9: This should be token, not root token

tests/unit/test_charm.py
gruyaume, line 1: The coverage in `src.charm.py` got from 84% to 75%. Are we testing stuff properly?

gruyaume (changes requested):

tests/unit/test_charm.py
Kayra1: Nope. Unit tests are coming up soon™

actions.yaml
Kayra1: changed

src/charm.py
Kayra1: nope

src/charm.py
Kayra1: this is a discussion I'd like to have with the team, I should have marked it as such. Instead, I'll remove it and create a ticket to address it later (this includes the above change with force_restore_snapshot). The gist of it is the charm will not be able to warn the user properly right now if they restore a vault that doesn't have the charm policy and role that was created when they first started vault. The options are either we handle that case, or we stop forcing snapshot restores.

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: this was for debugging. I don't think it matters. reversed it back.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -102,19 +102,31 @@ class VaultClientError(Exception):
...
         """Find and use the token related with the given auth method."""
-        auth_details.login(self._client)
+        try:
+            auth_details.login(self._client)
+        except (Forbidden, InternalServerError):
+            return False
+        return True
+
+    def is_authenticated(self) -> Optional[Dict]:
+        """Check if token is valid."""
```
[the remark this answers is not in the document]
Kayra1: fixed

src/charm.py
```diff
@@ -288,199 +305,34 @@ def _on_remove(self, event: RemoveEvent):
...
 
-    def _is_intermediate_ca_set(self, vault: Vault, common_name: str) -> bool:
-        """Check if the intermediate CA is set in the PKI secrets engine."""
-        intermediate_ca = vault.get_intermediate_ca(mount=PKI_MOUNT)
-        if not intermediate_ca:
-            return False
-        intermediate_ca_common_name = get_common_name_from_certificate(intermediate_ca)
-        return intermediate_ca_common_name == common_name
+        root_token = event.params.get("root-token", "")
+        v = Vault(self._api_address, self.tls.get_tls_file_path_in_charm(File.CA))
```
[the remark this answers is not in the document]
Kayra1: changed

src/charm.py
```diff
@@ -255,28 +277,23 @@ def _configure(self, event: Optional[ConfigChangedEvent] = None) -> None:  # noq
             )
 
     def _on_remove(self, event: RemoveEvent):
-        """Handle remove charm event.
+        """Handle certs removed event.
```
DanielArndt, line 280: This doesn't seem right. Isn't this listening to the event triggered by Juju on removal?

src/charm.py
```diff
@@ -288,199 +305,34 @@ def _on_remove(self, event: RemoveEvent):
                     pass
             self._delete_vault_data()
 
-    def _on_new_vault_kv_client_attached(self, event: NewVaultKvClientAttachedEvent):
-        """Handle vault-kv-client attached event."""
+    def _on_authorize_charm_action(self, event: ActionEvent) -> None:
         if not self.unit.is_leader():
-            logger.debug("Only leader unit can configure a vault-kv client, skipping")
+            event.fail("This action must be ran on the leader unit.")
```
DanielArndt, line 310:
```suggestion
            event.fail("This action must be run on the leader unit.")
```

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -102,19 +102,31 @@ class VaultClientError(Exception):
...
+            return None
+        return token_data
 
     def is_api_available(self) -> bool:
         """Return whether Vault is available."""
         try:
             self._client.sys.read_health_status(standby_ok=True)
             return True
-        except (VaultError, RequestException) as e:
+        except (VaultError, RequestException, InternalServerError) as e:
```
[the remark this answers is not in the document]
Kayra1: removed

src/charm.py
Kayra1: good catch, this is very old. reverted.

src/charm.py
Kayra1: fixed

lib/charms/vault_k8s/v0/vault_client.py
saltiyazan: But with this change now we won't be able to restore a Vault snapshot on a Vault cluster other than the one the snapshot was taken on in the first place. We won't be able to restore on a new deployment.
Force is used to allow restoring a snapshot of a Vault that use an unseal key X on a new Vault that uses an unseal key Y then unseal it with X once the snapshot is restored.
I'd not remove, but keep it and make sure it works.

tests/integration/test_integration.py
```diff
@@ -70,64 +70,137 @@ async def build_charms_and_deploy_vault(ops_test: OpsTest):
...
+        ops_test: OpsTest,
+        build_charms_and_deploy_vault: dict[str, Path | str],
+        initialize_leader_vault,
     ):
         assert ops_test.model
+        _, root_token, _ = initialize_leader_vault
         app: Application = ops_test.model.applications[APPLICATION_NAME]
-        await app.scale(NUM_VAULT_UNITS)
 
+        # TODO: maybe get the app ip and connect with cert?
```
saltiyazan, line 184: Should we keep this?

saltiyazan (changes requested):

lib/charms/vault_k8s/v0/vault_client.py
saltiyazan: I see it is not applied, never mind

src/charm.py
DanielArndt: Just going to leave my comments from a private conversation I had with Kayra:

I think it's reasonable for us to go into a blocked status when this happens. If they happen to restore a different Vault, or a Vault without the approle policy, we can force them to re-run the `authorize-charm` action which will either create or update the policy. They will need a token (likely the root token) to do this.

src/charm.py
```diff
@@ -288,199 +305,34 @@ def _on_remove(self, event: RemoveEvent):
...
-            logger.debug("Failed to sign the certificate")
-            return
-        self.vault_pki.set_relation_certificate(
-            relation_id=relation_id,
-            certificate=certificate.certificate,
-            certificate_signing_request=csr,
-            ca=certificate.ca,
-            chain=certificate.chain,
+        vault.configure_policy(policy_name=CHARM_POLICY_NAME, policy_path=CHARM_POLICY_PATH)
+        cidrs = [f"{self._bind_address}/24"]
```
gruyaume, line 327: This seems limiting, if the pod crash it will get a new IP, does this mean it won't have permissions anymore?

src/charm.py
```diff
@@ -288,199 +305,34 @@ def _on_remove(self, event: RemoveEvent):
...
+        role_id = vault.configure_approle(
+            role_name="charm",
+            cidrs=cidrs,
+            policies=[CHARM_POLICY_NAME, "default"],
         )
+        secret_id = vault.generate_role_secret_id(name="charm", cidrs=cidrs)
+        self._set_approle_auth_secret(role_id, secret_id)
+        self.on.config_changed.emit()
```
gruyaume, line 335: Why calling this here, this doesn't feel necessary

src/charm.py
```diff
@@ -288,199 +305,34 @@ def _on_remove(self, event: RemoveEvent):
...
-            return
-
-        relation = self.model.get_relation(event.relation_name, event.relation_id)
-
-        if relation is None or relation.app is None:
-            logger.warning(
-                "Relation or remote application is missing,"
-                "this should not happen, skipping event"
-            )
+        if "root" not in token_data["policies"]:
```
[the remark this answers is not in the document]
gruyaume: You may want to have all the api calls in a try/except (see the machine charm implementation) and fail the event if one of the calls fail.

gruyaume (changes requested):

tests/integration/test_integration.py
```diff
@@ -770,3 +898,72 @@ async def run_restore_backup_action(
...
+                "message": cells[4],
+            }
+        )
+    return output
+
+
+async def wait_for_vault_status_message(
+    ops_test: OpsTest, count: int, expected_message: str, timeout: int = 100, cadence: int = 5
+):
+    while timeout > 0:
```
gruyaume, line 929: There are two methods that do similar timeouts here, let's standardize with 1 approach (see `_get_vault_traefik_endpoint` which does it differently)

tests/integration/test_integration.py
```diff
@@ -770,3 +898,72 @@ async def run_restore_backup_action(
...
+
+
+def unseal_all_vaults(
+    ops_test: OpsTest, unit_addresses: List[str], root_token: str, unseal_key: str
+):
+    for endpoint in unit_addresses:
+        unseal_vault(endpoint, root_token, unseal_key)
+
+
+async def authorize_charm(ops_test: OpsTest, root_token: str):
```
gruyaume, line 958: Add return typing

tests/integration/test_integration.py
```diff
@@ -770,3 +898,72 @@ async def run_restore_backup_action(
 
 def copy_lib_content() -> None:
     shutil.copyfile(src=VAULT_KV_LIB_DIR, dst=f"{VAULT_KV_REQUIRER_CHARM_DIR}/{VAULT_KV_LIB_DIR}")
+
+
+async def read_vault_unit_statuses(ops_test: OpsTest):
```
gruyaume, line 903: Add return typing

tests/integration/test_integration.py
```diff
@@ -673,6 +795,12 @@ async def run_get_ca_certificate_action(ops_test: OpsTest, timeout: int = 60) ->
     return await ops_test.model.get_action_output(action_uuid=action.entity_id, wait=timeout)
 
 
+async def get_vault_ca_certificate(vault_unit: Unit) -> str:
```
gruyaume, line 798: Can't we use the juju secrets like we do in documentation?

tests/integration/test_integration.py
```diff
@@ -70,64 +70,137 @@ async def build_charms_and_deploy_vault(ops_test: OpsTest):
...
+        status="blocked",
         timeout=1000,
         wait_for_exact_units=NUM_VAULT_UNITS,
     )
 
     return {"vault-kv-requirer": vault_kv_requirer_charm}
 
 
+@pytest.fixture(scope="module")
+async def initialize_leader_vault(ops_test: OpsTest, build_charms_and_deploy_vault):
```
gruyaume, line 82: Add return typing

tests/integration/test_integration.py
```diff
@@ -45,11 +45,11 @@
 }
 
 NUM_VAULT_UNITS = 3
+VAULT_UNIT_DOMAIN = "vault-k8s-endpoints.vault.svc.cluster.local"
```
gruyaume, line 48: Where is this used?

gruyaume (changes requested):

gruyaume:
Also, please improve the PR description "This change does a bunch of things" isn't all that helpful. This is a change in user experience, you may want to explain what we are doing precisely and how it impacts users.

src/charm.py
DanielArndt: Related: https://github.com/canonical/vault-operator/pull/62#discussion_r1516346734

I don't think we can restrict this, as we don't know where Juju is going to place the new pod.

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: We're definitely gonna support that. I'm resolving this thread since this is outdated.

src/charm.py
Kayra1: The functionality in the comments is here now.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -102,12 +102,24 @@ class VaultClientError(Exception):
...
+    def __init__(self, url: str, ca_cert_path: Optional[str]):
+        self._client = hvac.Client(url=url, verify=ca_cert_path if ca_cert_path else False)
 
-    def authenticate(self, auth_details: AuthMethod) -> None:
+    def authenticate(self, auth_details: AuthMethod) -> bool:
         """Find and use the token related with the given auth method."""
-        auth_details.login(self._client)
+        try:
+            auth_details.login(self._client)
+        except (Forbidden, InternalServerError):
```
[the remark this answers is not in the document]
Kayra1: Done

src/charm.py
```diff
@@ -186,24 +193,35 @@ def _on_collect_status(self, event: CollectStatusEvent):  # noqa: C901
...
-            event.add_status(WaitingStatus("Waiting for CA certificate in workload"))
+        if not self.unit.is_leader() and not self.tls.tls_file_pushed_to_workload(File.CA):
+            event.add_status(WaitingStatus("Waiting for CA certificate to be shared"))
             return
-        if not self.tls.tls_file_available_in_charm(File.CA):
-            event.add_status(WaitingStatus("Waiting for CA certificate in charm"))
+        try:
+            self.tls.get_tls_file_path_in_charm(File.CA)
+        except VaultCertsError:
+            event.add_status(BlockedStatus("Missing Storage"))
```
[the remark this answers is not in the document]
Kayra1: I changed it.

For context, this runs only once in the charms entire lifetime, on storage detached but before remove. Most users will never see it.

src/charm.py
```diff
@@ -186,24 +193,35 @@ def _on_collect_status(self, event: CollectStatusEvent):  # noqa: C901
...
             return
         vault = Vault(
             url=self._api_address, ca_cert_path=self.tls.get_tls_file_path_in_charm(File.CA)
         )
         if not vault.is_api_available():
             event.add_status(WaitingStatus("Waiting for vault to be available"))
             return
+        if not vault.is_initialized():
+            event.add_status(BlockedStatus("Waiting for vault to be initialized"))
```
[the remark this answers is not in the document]
Kayra1: agree. changed. These are the exact strings in https://github.com/canonical/vault-operator/pull/62, maybe opening a PR there to change them too makes sense after this.

src/charm.py
```diff
@@ -186,24 +193,35 @@ def _on_collect_status(self, event: CollectStatusEvent):  # noqa: C901
...
+            return
+        if vault.is_sealed():
+            event.add_status(BlockedStatus("Waiting for vault to be unsealed"))
+            return
+        role_id, secret_id = self._get_approle_auth_secret()
+        if not role_id or not secret_id:
+            event.add_status(BlockedStatus("Waiting for charm to be authorized"))
```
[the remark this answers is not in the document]
Kayra1: see above

src/charm.py
```diff
@@ -186,24 +193,35 @@ def _on_collect_status(self, event: CollectStatusEvent):  # noqa: C901
...
+            return
+        if not self._get_active_vault_client():
+            event.add_status(BlockedStatus("Waiting for a leader unit to be chosen"))
```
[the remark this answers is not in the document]
Kayra1: No, I changed it.

src/charm.py
```diff
@@ -186,24 +193,35 @@ def _on_collect_status(self, event: CollectStatusEvent):  # noqa: C901
...
             return
-        if not self.tls.tls_file_pushed_to_workload(File.CA):
-            event.add_status(WaitingStatus("Waiting for CA certificate in workload"))
+        if not self.unit.is_leader() and not self.tls.tls_file_pushed_to_workload(File.CA):
+            event.add_status(WaitingStatus("Waiting for CA certificate to be shared"))
             return
-        if not self.tls.tls_file_available_in_charm(File.CA):
-            event.add_status(WaitingStatus("Waiting for CA certificate in charm"))
+        try:
+            self.tls.get_tls_file_path_in_charm(File.CA)
```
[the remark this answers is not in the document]
Kayra1: No. I said this above too, but this function will error out only once, after storage detached. But for consistency I moved it.

src/charm.py
```diff
@@ -262,21 +284,16 @@ def _on_remove(self, event: RemoveEvent):
...
-            if self._bind_address:
-                vault = Vault(
-                    url=self._api_address,
-                    ca_cert_path=self.tls.get_tls_file_path_in_charm(File.CA),
-                )
-                vault.authenticate(Token(root_token))
+            role_id, secret_id = self._get_approle_auth_secret()
+            if role_id and secret_id and self._bind_address:
+                vault = Vault(url=self._api_address, ca_cert_path=None)
+                vault.authenticate(AppRole(role_id, secret_id))
```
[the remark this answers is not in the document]
Kayra1: yes, we can't using a CA here.

src/charm.py
```diff
@@ -262,21 +284,16 @@ def _on_remove(self, event: RemoveEvent):
         if not self._container.can_connect():
             return
         try:
-            root_token, unseal_keys = self._get_initialization_secret()
-            if self._bind_address:
-                vault = Vault(
-                    url=self._api_address,
-                    ca_cert_path=self.tls.get_tls_file_path_in_charm(File.CA),
```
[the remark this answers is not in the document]
Kayra1: I think so

src/charm.py
Kayra1: No, there is a pod crashing test that already passes with this, and the CIDR is contained to the subnet that k8s assigns the namespace. In k8s, this will always without exception work.

Kayra1:
> Also, please improve the PR description "This change does a bunch of things" isn't all that helpful. This is a change in user experience, you may want to explain what we are doing precisely and how it impacts users.

Copied over Dan's description on the machine charm PR.

src/charm.py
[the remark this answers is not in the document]
Kayra1: Changed.

src/charm.py
```diff
@@ -288,199 +305,34 @@ def _on_remove(self, event: RemoveEvent):
...
-            certificate=certificate.certificate,
-            certificate_signing_request=csr,
-            ca=certificate.ca,
-            chain=certificate.chain,
+        vault.configure_policy(policy_name=CHARM_POLICY_NAME, policy_path=CHARM_POLICY_PATH)
+        cidrs = [f"{self._bind_address}/24"]
+        role_id = vault.configure_approle(
+            role_name="charm",
+            cidrs=cidrs,
+            policies=[CHARM_POLICY_NAME, "default"],
```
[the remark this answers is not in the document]
Kayra1: I don't understand the error path here. If configure policy or approle fails, this part can't run in configure anyway. If the approle secret was set, this part needs to have succeeded, otherwise the approle would be wrong and the action would need to be ran again anyway. If the action fails even without a fail by vault not accepting any of the vault functions, the action will need to be ran again anyway.

This was also exactly how the machine charm did it.

src/charm.py
```diff
@@ -602,8 +454,6 @@ def _on_restore_backup_action(self, event: ActionEvent) -> None:
 
         Restores the snapshot with the provided ID.
         Unseals Vault using the provided unseal key.
-        Sets the root token to the provided root token.
```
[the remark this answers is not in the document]
Kayra1: removed.

src/charm.py
Kayra1: Dan't right probably, removed it.

tests/integration/test_integration.py
Kayra1: We can, but we would have to have complex logic to check if it's coming from `self-signed-certificates` or vault generated. Also wouldn't work to validate the result of `tls-certificates-access` integration since we'd be checking the source and not the end result. This is way simpler and always gives us the live CA that vault sees.

tests/integration/test_integration.py
Kayra1: at some point it was used to build the requests to the vault units, but i started extracting those from the status. Removed.

tests/integration/test_integration.py
Kayra1: Done

tests/integration/test_integration.py
Kayra1: Done

tests/integration/test_integration.py
Kayra1: Done

tests/integration/test_integration.py
Kayra1: Didn't even notice that. Moved them closer together and updated both, also added docstring.

tests/integration/test_integration.py
Kayra1: i'll just remove it this pr is already large enough.

DanielArndt:
https://github.com/canonical/vault-k8s-operator/pull/142/commits/3c495e5e4376a09eab0b5e1376d2f1ad9a2ffb8f has no functional changes, I just put everything that was moved back to where it was before.

That should make keeping up with main easier, and focus the PR on the changes.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
saltiyazan: Wouldn't it be useful to log the actual exception too? This would be valid for the whole lib.

src/charm.py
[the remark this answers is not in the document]
saltiyazan: I still don't get the message itself.

src/charm.py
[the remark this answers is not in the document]
saltiyazan: Why not?

src/templates/charm_policy.hcl
```diff
@@ -0,0 +1,66 @@
+# Allow management of policies starting with charm- prefix
+path "sys/policy/charm-*" {
+  capabilities = ["create", "read", "update", "delete"]
+}
```
saltiyazan, line 4: in `[_ensure_unit_credentials](https://github.com/canonical/vault-k8s-operator/blob/main/src/charm.py#L787)` we configure a policy that doesn't use this prefix.
Will that be a problem?

src/templates/charm_policy.hcl
```diff
@@ -0,0 +1,66 @@
...
+  capabilities = ["list"]
+}
+
+# Allow charm- prefixes secrets backends to be mounted and managed
+path "sys/mounts/charm-*" {
+  capabilities = ["create", "read", "update", "delete", "sudo"]
+}
+
+# Allow charm- prefixes pki backends to be used
+path "charm-pki-*" {
```
saltiyazan, line 34: Isn't this included in the one specific for `charm-*` a few line above?

src/templates/charm_policy.hcl
```diff
@@ -0,0 +1,66 @@
...
+}
+
+# Allow discovery of secrets backends
+path "sys/mounts" {
+  capabilities = ["read"]
+}
+
+path "sys/mounts/" {
+  capabilities = ["list"]
+}
```
saltiyazan, line 45:
```suggestion
path "sys/mounts" {
  capabilities = ["read", "list"]
}

```

src/templates/charm_policy.hcl
```diff
@@ -0,0 +1,66 @@
...
+  capabilities = ["create", "read", "update", "delete", "list"]
+}
+
+# Allow discovery of approles
+path "auth/approle/role" {
+  capabilities = ["read"]
+}
+path "auth/approle/role/" {
+  capabilities = ["list"]
+}
```
saltiyazan, line 26:
```suggestion
path "auth/approle/role" {
  capabilities = ["read", "list"]
}

```

src/templates/charm_policy.hcl
Kayra1: list endpoints need trailing slashes.

src/templates/charm_policy.hcl
Kayra1: same as above.

src/charm.py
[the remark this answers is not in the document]
Kayra1: Because the storage for certificates are removed before _on_remove fires.

tests/unit/test_charm.py
Kayra1: Me and dan added a bunch of unit tests. Now, the percentage difference between main and this branch is 1%.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
Kayra1: I added that here, but I don't think it would be useful to add it to the ones where we already check what the message says. Forbidden just returns a really generic permission denied string, and the already enabled exceptions also say very generic things like request failed.

src/charm.py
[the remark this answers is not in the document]
Kayra1: Changed again

src/templates/charm_policy.hcl
Kayra1: which policy would that be? We only configure policies that start with charm [here](https://github.com/canonical/vault-k8s-operator/blob/058889f586c4e440a4cb502fa0d8af11ecb8eb85/src/charm.py#L426).

src/templates/charm_policy.hcl
Kayra1: You're right. I've removed this one.

tests/unit/test_charm.py
```diff
@@ -264,56 +244,107 @@ def test_given_bind_address_not_available_when_evaluate_status_then_status_is_wa
             WaitingStatus("Waiting for bind and ingress addresses to be available"),
         )
 
+    @patch("charm.Vault", autospec=True)
     @patch("ops.model.Model.get_binding")
-    def test_given_not_leader_and_init_secret_not_set_when_evaluate_status_then_status_is_waiting(
-        self, patch_get_binding
+    def test_given_vault_uninitialized_when_evaluate_status_then_status_is_blocked(
```
saltiyazan, line 249: Not a critical thing, but we might want to use either the prefix mock or the prefix patch for consistency, instead of mixing both. This has been the case before too so I don't expect all names to get aligned in this PR :)

tests/unit/test_charm.py
```diff
@@ -264,56 +244,107 @@ def test_given_bind_address_not_available_when_evaluate_status_then_status_is_wa
...
+        self, mock_get_binding: MagicMock, mock_vault_class: MagicMock
```
saltiyazan, line 250: Not a critical thing, but we might want to use either the prefix mock or the prefix patch for consistency, instead of mixing both. This has been the case before too so I don't expect all names to get aligned in this PR :)

Kayra1: the patch decorator replaces a real object with a different object, which in our case is always a mock object. So we have to either call everything a patch or everything a mock.

Kayra1: I can mass replace all of them pretty easily with vim, but i think it makes sense to do it in a different PR since this one is very large.

tests/unit/test_charm.py
saltiyazan, line 1: Tests look good in general.
I am only missing test cases for the collect status when the charm would be waiting for the CA certificates.

tests/unit/test_charm.py
saltiyazan: I agree, it's fine yo have it later

tests/unit/test_charm.py
Kayra1: added a couple more tests

src/charm.py
```diff
@@ -669,79 +718,29 @@ def _on_restore_backup_action(self, event: ActionEvent) -> None:
...
+        try:
+            if self._approle_secret_set():
+                role_id, secret_id = self._get_approle_auth_secret()
+                vault = Vault(
+                    url=self._api_address,
+                    ca_cert_path=self.tls.get_tls_file_path_in_charm(File.CA),
+                )
+                if role_id and secret_id and not vault.authenticate(AppRole(role_id, secret_id)):
+                    self._remove_approle_auth_secret()
+        except Exception as e:
```
DanielArndt, line 739: Do we really want to catch a bare `Exception` here? this seems like a bad idea

tests/integration/test_integration.py
```diff
@@ -67,66 +67,143 @@ async def build_charms_and_deploy_vault(ops_test: OpsTest):
...
 
     return {"vault-kv-requirer": vault_kv_requirer_charm}
 
 
+@pytest.fixture(scope="module")
+async def initialize_leader_vault(
+    ops_test: OpsTest, build_charms_and_deploy_vault: Dict[str, Path | str]
+) -> Tuple[int, str, str]:
+    leader_unit = await get_leader_unit(ops_test.model, APPLICATION_NAME)
+    leader_unit_index = int(leader_unit.name[-1])
```
DanielArndt, line 83: This doesn't seem very future proof. What if the unit isn't a single digit? Maybe split on `/` first?

DanielArndt (commented):
A few comments, but really the only deal-breaker for me is the bare `Exception` catch.

Otherwise I think this looks good.

DanielArndt (changes requested):
Meant to submit last one as "request changes"

src/charm.py
Kayra1: If we hadn't, we would have missed the http Exception about rewinding the body. But I recognize that catching those is a pandoras box so I can change this into VaultClientError and let pythons normal stack trace passthrough on rare exceptions like that.

tests/integration/test_integration.py
Kayra1: I *highly* doubt the gh runners can handle more than 10 pods, but splitting on `/` is just better so I changed it.

src/charm.py
DanielArndt: I think it's actually desirable for our charm to crash when unexpected errors occur, so I much prefer this.

gruyaume (approved):

saltiyazan (approved):

PR #153, opened 2024-02-07
chore: Adds substrate abstraction in vault_tls

saltiyazan:
# Description

The aim of this PR is to improve the `vault_tls` lib so it can be used in the machine charm too.
It adds `substrate` abstraction and creates the Container class.
Uses V3 of the tls certificates lib.

This PR does not include refactoring of the lib or change of the design for now.

# Checklist:


lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -222,9 +248,9 @@ def _should_request_new_certificate(self) -> bool:
         pending_csrs = self.tls_access.get_certificate_signing_requests(unfulfilled_only=True)
         expired_certs = self.tls_access.get_expiring_certificates()
 
-        existing_csr_is_fulfilled = any([existing_csr in csr.values() for csr in fulfilled_csrs])
-        existing_csr_is_pending = any([existing_csr in csr.values() for csr in pending_csrs])
-        existing_csr_expiring = any([existing_csr in cert.values() for cert in expired_certs])
+        existing_csr_is_fulfilled = any([existing_csr in csr.csr for csr in fulfilled_csrs])
+        existing_csr_is_pending = any([existing_csr in csr.csr for csr in pending_csrs])
+        existing_csr_expiring = any([existing_csr in cert.certificate for cert in expired_certs])
```
Kayra1, line 253: `csr.csr` reads weird here. maybe replace replace the name of the `csr` object with `object` or `entry` or something like that? It's not that important though.

Kayra1 (changes requested):

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -123,42 +142,43 @@ def _on_tls_certificates_access_relation_broken(self, event: RelationBrokenEvent
         and reloads vault.
         """
         self._remove_all_certs_from_workload()
-        if not self.charm._ingress_address:
+        if not self.subject_ip:
+            event.defer()
```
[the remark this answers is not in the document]
saltiyazan: in that case it should be enough to have only the following lines:
```
self._remove_all_certs_from_workload()
self.charm.on.config_changed.emit()
```
no? and then there is no need to have knowledege of `subject_ip` as it is a parameter everywhere else.
What do you think?

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -46,27 +47,49 @@ def process(self, msg, kwargs):
...
-TLS_FILE_FOLDER_PATH = "/vault/certs"
 CA_CERTIFICATE_JUJU_SECRET_KEY = "vault-ca-certificates-secret-id"
 CA_CERTIFICATE_JUJU_SECRET_LABEL = "vault-ca-certificate"
 
 VAULT_CA_SUBJECT = "Vault self signed CA"
 
 
-class Substrate(Enum):
-    """Determines which type of charm this library is running in.
+class Substrate(ABC):
```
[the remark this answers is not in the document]
saltiyazan: would something like `WorkloadHost ` work? I don't have many better ideas neither.

src/charm.py
```diff
@@ -65,6 +61,7 @@
 S3_RELATION_NAME = "s3-parameters"
 REQUIRED_S3_PARAMETERS = ["bucket", "access-key", "secret-key", "endpoint"]
 BACKUP_KEY_PREFIX = "vault-backup"
+TLS_FILE_FOLDER_PATH = "/vault/certs"
```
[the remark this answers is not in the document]
saltiyazan: The issue with this is that it's "substrate" specific.
in case of machine it would be "/var/snap/vault/common/certs"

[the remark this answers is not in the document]
Kayra1: Oh ok, that makes sense. Then can we name it that indicates that, like `container_tls_folder_path`?

lib/charms/vault_k8s/v0/vault_tls.py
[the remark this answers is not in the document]
saltiyazan: I mean we emit a config_changed event and leave it to the charm to run `_configure`.
the ip should be available there in the charm, if it isn't available there we there will be an error and I think that's fine, until it is available again and `_configure` runs again.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -342,23 +362,20 @@ def ca_certificate_set_in_peer_relation(self) -> bool:
...
-        tlslogger.debug("Removed existing certificate files from workload.")
+        tls_logger.debug("Removed existing certificate files from workload.")
 
-    def _reload_vault_container(self) -> None:
+    def _reload_vault(self) -> None:
         """Sends a SIGHUP signal to the container which reloads Vault's files. Fails gracefully."""
         try:
-            self.charm._container.send_signal(SIGHUP, self.charm._container_name)
-            tlslogger.debug("Container restart requested")
+            self.substrate.send_signal(SIGHUP, "vault")
```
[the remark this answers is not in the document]
saltiyazan: > I still see no issue in using variables from the charm. That's the whole point of passing self as an argument to this class.
anything we access from the charm will be a requirement that we need to fulfil by the charm.
I prefer having the charm pass what it is required by lib in the constructor, defining a clear interface event if that means we will have a large init.
Passing self should be to access things like `charm.model` and `charm.app`.
At least that's how I have seen it done in other libs with a few exceptions. Of course we can have some exceptions here too as this lib is not expected to be used by charms other that `vault`, but I prefer to limit those exceptions to functions calls and not variables/constants.

src/charm.py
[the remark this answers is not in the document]
saltiyazan: done

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: done

lib/charms/vault_k8s/v0/vault_tls.py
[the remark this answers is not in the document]
saltiyazan: For now I will be passing this as a parameter in the constructor.
I'm happy to change it to use a getter function form the charm if that sounds better.

lib/charms/vault_k8s/v0/vault_tls.py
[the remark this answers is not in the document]
Kayra1: I'll leave it up to you.

lib/charms/vault_k8s/v0/vault_tls.py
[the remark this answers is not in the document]
Kayra1: > Passing self should be to access things like charm.model and charm.app

I don't see how this is any different. Why not pass `model` and `app` as init arguments also then? The container name is just as native of a value as `model` or `app` to the charm. If you want to absolutely guarantee that this value is not stale, you can locate the name of the container from the pebble layer too. A getter function just wraps this variable with extra lines, so I'd rather not do that.

I'm ok with passing it in the init. I'd like to prevent this seemingly unrelated function call from failing when the name of the container in the pebble layer changes when there's really no need for it to do that, even if I don't expect that to ever happen.

lib/charms/vault_k8s/v0/vault_tls.py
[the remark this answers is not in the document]
Kayra1: I remember that this didn't work when I was trying locally, but that could have been many review comments and changes ago. Try doing your suggested 2 line change. If you manually relate and unrelate a certificates provider locally and there seems to be no issue and all the tests are passing, then that's the better option and we should go with that.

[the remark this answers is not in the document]
saltiyazan: I tested it and it works.

src/charm.py
```diff
@@ -146,6 +144,7 @@ def _configure(self, event: Optional[ConfigChangedEvent] = None) -> None:
                 "Waiting for bind and ingress addresses to be available"
             )
             return
+        self.tls.subject_ip = self._ingress_address  # type: ignore[assignment]
```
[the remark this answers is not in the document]
saltiyazan: This not needed anymore

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -81,14 +98,16 @@ class File(Enum):
...
+    def __init__(self, charm, peer_relation: str, substrate: Substrate, tls_folder_path: str):
         """Manager of TLS relation and configuration."""
         super().__init__(charm, "tls")
         self.charm = charm
         self.substrate = substrate
         self.peer_relation = peer_relation
+        self.subject_ip = None
         self.tls_access = TLSCertificatesRequiresV2(charm, TLS_CERTIFICATE_ACCESS_RELATION_NAME)
         self.certificate_transfer = CertificateTransferProvides(charm, SEND_CA_CERT_RELATION_NAME)
+        self.tls_folder_path = tls_folder_path
```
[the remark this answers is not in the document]
saltiyazan: I agree with you, but `vault_tls` doesn't really know about that.
We either access the constant directly from the charm or we pass it in the constructor.
I chose passing it.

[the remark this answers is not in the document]
Kayra1: Looks good right now

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -99,7 +130,8 @@ def __init__(self, charm, peer_relation: str, substrate: Substrate):
             self._on_tls_certificates_access_relation_broken,
         )
         self.framework.observe(
-            self.tls_access.on.certificate_available, self._on_certificate_config_changed
+            self.tls_access.on.certificate_available,
+            self._on_tls_certificates_access_certificate_available,
```
Kayra1, line 134: We have decided against listening to this event and centralizing the logic in the configure function in the first implementation of this library, so this should be reversed.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -123,56 +155,53 @@ def _on_tls_certificates_access_relation_broken(self, event: RelationBrokenEvent
...
-                self._push_tls_file_to_workload(File.CA, assigned_cert["ca"])
-                tlslogger.info(
-                    "Certificate from access relation saved for unit %s.",
-                    self.charm.unit.name,
-                )
-                self._reload_vault_container()
-                return
+            self._reload_vault()
+            return
+        return
```
Kayra1, line 204: See comment above

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -217,39 +246,46 @@ def _should_request_new_certificate(self) -> bool:
...
-        existing_csr_expiring = any([existing_csr in cert.values() for cert in expired_certs])
+        if existing_csr_is_fulfilled and not existing_csr_expiring:
+            return False
 
-        if csr_is_in_workload:
-            if existing_csr_is_fulfilled and not existing_csr_expiring:
-                return False
-            if existing_csr_is_pending:
-                return False
         return True
```
Kayra1, line 270: I think grouping up the if statements at the end makes this more readable than how it is here personally, but it looks like the logic is still the same so it's not that important.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -279,101 +315,80 @@ def get_tls_file_path_in_charm(self, file: File) -> str:
...
             "privatekey": private_key,
             "certificate": certificate,
         }
-        juju_secret = self.charm.app.add_secret(
-            juju_secret_content, label=CA_CERTIFICATE_JUJU_SECRET_LABEL
-        )
-        peer_relation = self.charm.model.get_relation(self.peer_relation)
-        peer_relation.data[self.charm.app].update({CA_CERTIFICATE_JUJU_SECRET_KEY: juju_secret.id})
-        tlslogger.debug("Vault CA certificate secret set in peer relation")
+        self.charm.app.add_secret(juju_secret_content, label=CA_CERTIFICATE_JUJU_SECRET_LABEL)
```
Kayra1, line 346: Is there any difference or issue using `add_secret` over `set_content` except for the creation of the secret? Any problem with revisions or anything?

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -279,101 +315,80 @@ def get_tls_file_path_in_charm(self, file: File) -> str:
...
-    def ca_certificate_set_in_peer_relation(self) -> bool:
-        """Returns whether CA certificate is stored in peer relation data."""
+    def ca_certificate_secret_set(self) -> bool:
+        """Returns whether CA certificate is stored in secret."""
         try:
-            ca_private_key, ca_certificate = self._get_ca_certificate_secret_in_peer_relation()
+            ca_private_key, ca_certificate = self._get_ca_certificate_secret()
             if ca_private_key and ca_certificate:
                 return True
         except PeerSecretError:
```
Kayra1, line 355: This exception should also be renamed or just use SecretNotFoundError.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -279,101 +315,80 @@ def get_tls_file_path_in_charm(self, file: File) -> str:
...
-        if not self.charm._is_peer_relation_created():
-            raise RuntimeError("Peer relation not created")
-        peer_relation = self.model.get_relation(self.peer_relation)
-        if peer_relation:
-            peer_relation.data[self.charm.app].pop(CA_CERTIFICATE_JUJU_SECRET_KEY)
-            tlslogger.debug("Vault CA certificate secret removed from peer relation")
-
-    def ca_certificate_set_in_peer_relation(self) -> bool:
-        """Returns whether CA certificate is stored in peer relation data."""
+    def ca_certificate_secret_set(self) -> bool:
```
Kayra1, line 349: We already use the `set` verb when setting it
```suggestion
    def ca_certificate_secret_exists(self) -> bool:
```

Kayra1 (changes requested):
Great job removing the peer relation from everywhere, the lib's looking much better.

gruyaume (approved):

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: I agree, it was slightly more readable
but here we return early if it's the case, and readability has not gone that bad.
I don't insist on keeping it tho
let me know

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: I think this will not support updating the secret. We don't modify it right now, but it might be something we need to consider, but I'd leave it out of this PR and keep as it has been.

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: Good point, fixed

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: Done

lib/charms/vault_k8s/v0/vault_tls.py
Kayra1: maybe a quick `if exists set content`? As long as the behavior is documented i don't mind having either.

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: Done.

lib/charms/vault_k8s/v0/vault_tls.py
saltiyazan: Done

lib/charms/vault_k8s/v0/vault_tls.py
Kayra1: I don't see not returning early here as too much of a big deal since we only expect vault to have a single CSR in this relation databag. I'd prefer to revert it.

saltiyazan: Alright, done.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -123,56 +164,49 @@ def _on_tls_certificates_access_relation_broken(self, event: RelationBrokenEvent
...
+            ) or not self._tls_file_pushed_to_workload(File.CERT):
+                self._generate_self_signed_certs(subject_ip)
+                tls_logger.info(
+                    "Saved Vault generated CA and self signed certificate to %s.",
+                    self.charm.unit.name,
+                )
+                self._reload_vault()
 
-        elif self.charm.model.get_relation(TLS_CERTIFICATE_ACCESS_RELATION_NAME):
+        else:
```
Kayra1, line 190: You can just remove the else here and return after reloading since we don't do anything after the else.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -123,56 +164,49 @@ def _on_tls_certificates_access_relation_broken(self, event: RelationBrokenEvent
...
+                self._push_tls_file_to_workload(File.CERT, assigned_cert.certificate)
+                self._push_tls_file_to_workload(File.CA, assigned_cert.ca)
+                tls_logger.info(
                     "Certificate from access relation saved for unit %s.",
                     self.charm.unit.name,
                 )
-                self._reload_vault_container()
+                self._reload_vault()
                 return
+            return
```
Kayra1, line 209: No need for either of these

Kayra1 (approved):
Great job. Just some minor comments.

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -46,27 +48,49 @@ def process(self, msg, kwargs):
...
-CA_CERTIFICATE_JUJU_SECRET_KEY = "vault-ca-certificates-secret-id"
-CA_CERTIFICATE_JUJU_SECRET_LABEL = "vault-ca-certificate"
+CA_CERTIFICATE_JUJU_SECRET_LABEL = "self-signed-vault-ca-certificate"
 
 VAULT_CA_SUBJECT = "Vault self signed CA"
 
 
-class Substrate(Enum):
-    """Determines which type of charm this library is running in.
+# TODO Move this class, it doesn't belong here.
```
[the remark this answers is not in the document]
saltiyazan: This should be addressed very soon once the spec around charm operation generalisation is accepted, that's why I was comfortable adding this here.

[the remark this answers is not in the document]
Kayra1: I like leaving them. I have an extension that lists them and i can see that some of the libraries we use also have them merged and uploaded

gruyaume (approved):

PR #176, opened 2024-02-16
chore: Uses a pyproject.toml file that matches the one in the machine charm

saltiyazan:
# Description

As we are going towards sharing code between the machine and the k8s charms, I think it might be a good idea to have the same rules and tools used in pyproject.toml for both.

- ruff is added and it replaces isort too.

# Checklist:


pyproject.toml
```diff
@@ -1,13 +1,40 @@
...
+minversion = "6.0"
+log_cli_level = "INFO"
+
+# Formatting tools configuration
 [tool.black]
 line-length = 99
 target-version = ["py38"]
 
+# Linting tools configuration
+[tool.ruff]
```
gruyaume, line 18: We're adding ruff related validation but we're not using ruff in `tox`. Make sure we completely make the switch to ruff here.

saltiyazan: Done, please note that `ruff` replaces `isort` added in PR description too.

gruyaume (approved):

PR #183, opened 2024-02-23
chore: centralize status management

gruyaume:
# Description

Centralize status management using the Collect Status event handler from ops. 

## Rationale

As we are adding more event handlers in the context of PKI, it became clear that having central status management would simplify operations. Specifically, without this change here, we would have had to move the charm to blocked status in the `RelationJoined` event in cases where the `common_name` config is not set. That would have been an issue as a follow-up `update_status` event would have overwritten this status. 

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_tls.py
+++ b/lib/charms/vault_k8s/v0/vault_tls.py
@@ -181,9 +181,9 @@ class VaultTLSManager(Object):
                 ca_private_key, ca_certificate = generate_vault_ca_certificate()
                 self._set_ca_certificate_secret(ca_private_key, ca_certificate)
                 tls_logger.info("Saved the Vault generated CA cert in juju secrets.")
-            if not self._tls_file_pushed_to_workload(
+            if not self.tls_file_pushed_to_workload(
                 File.CA
-            ) or not self._tls_file_pushed_to_workload(File.CERT):
+            ) or not self.tls_file_pushed_to_workload(File.CERT):
                 self._generate_self_signed_certs(subject_ip)
                 tls_logger.info(
                     "Saved Vault generated CA and self signed certificate to %s.",
@@ -361,7 +361,7 @@ class VaultTLSManager(Object):
 
     def ca_certificate_is_saved(self) -> bool:
         """Return wether a CA cert is saved in the charm."""
-        return self.ca_certificate_secret_exists() or self._tls_file_pushed_to_workload(File.CA)
+        return self.ca_certificate_secret_exists() or self.tls_file_pushed_to_workload(File.CA)
 
     def _remove_all_certs_from_workload(self) -> None:
         """Remove the certificate files that are used for authentication."""
@@ -421,7 +421,7 @@ class VaultTLSManager(Object):
             pass
         tls_logger.debug("Removed %s file from workload.", file.name)
 
-    def _tls_file_pushed_to_workload(self, file: File) -> bool:
+    def tls_file_pushed_to_workload(self, file: File) -> bool:
         """Return whether tls file is pushed to the workload.
 
         Args:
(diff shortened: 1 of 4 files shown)


lib/charms/vault_k8s/v0/vault_tls.py
Kayra1, line 1: Bump version pls

src/charm.py
```diff
@@ -142,38 +145,73 @@ def __init__(self, *args):
     def _on_install(self, event: InstallEvent):
         """Handle the install charm event."""
         if not self._container.can_connect():
-            self.unit.status = WaitingStatus("Waiting to be able to connect to vault unit")
             event.defer()
             return
         self._delete_vault_data()
 
+    def _on_collect_status(self, event: CollectStatusEvent):
```
Kayra1, line 152: I just find this paradigm so much more confusing than keeping everything in configure. How would we even make sure to keep the order of things in sync between configure and collect status? 

How about checking if the event is a collect status event in configure and adding the status only if? How about not using collect status at all? idk

tests/unit/lib/charms/vault_k8s/v0/test_vault_tls.py
```diff
@@ -132,11 +132,11 @@ def test_given_not_leader_and_ca_not_set_when_configure_then_status_is_waiting(
             bind_address="1.2.1.2", ingress_address="10.1.0.1"
         )
 
-        self.harness.charm.on.config_changed.emit()
+        self.harness.evaluate_status()
```
Kayra1, line 135: How about running both functions? I feel like we should check that configure returns properly too.

Applies to all the other changes below too.

Kayra1 (changes requested):
Some comments

lib/charms/vault_k8s/v0/vault_tls.py
```diff
@@ -317,6 +317,20 @@ def get_tls_file_path_in_charm(self, file: File) -> str:
         storage_location = cert_storage.location
         return f"{storage_location}/{file.name.lower()}.pem"
 
+    def tls_file_available_in_charm(self, file: File) -> bool:
```
saltiyazan, line 320: How is the outcome of this function different to `tls_file_pushed_to_workload`?

src/charm.py
```diff
@@ -142,38 +145,73 @@ def __init__(self, *args):
...
+            return
+        if not self.unit.is_leader() and not self._initialization_secret_set_in_peer_relation():
+            event.add_status(WaitingStatus("Waiting for initialization secret"))
+            return
+        if not self.tls.tls_file_pushed_to_workload(File.CA):
+            event.add_status(WaitingStatus("Waiting for TLS certificates"))
+            return
+        if not self.tls.tls_file_available_in_charm(File.CA):
+            event.add_status(WaitingStatus("Waiting for CA certificate"))
+            return
```
saltiyazan, line 189: I'm not sure I understand the difference between the two.

lib/charms/vault_k8s/v0/vault_tls.py
gruyaume: I answered this in the other question below.

src/charm.py
gruyaume: On storage detaching event, it is possible for one of those conditions to be true and not the other (i.e. the tls file can be in the workload but not available in the charm), hence the need for the 2. I improved the content of the status message to make it a bit clearer.

lib/charms/vault_k8s/v0/vault_tls.py
gruyaume: Done

src/charm.py
gruyaume: There is no need to keep anything in sync between configure and collect status. Status are now **only** managed in collect status. collect status is fired at the end of every event making it so that we don't have to maintain status anywhere else. 

As mentioned in the description, the alternative is to manage status in multiple event handlers which is much worse.

tests/unit/lib/charms/vault_k8s/v0/test_vault_tls.py
gruyaume: I only changed the tests where we were validating that the status was the right one. Since status is only maintained by collect status, I see no need to also call another event (ex. config_changed)

saltiyazan (approved):

Kayra1 (approved):

PR #184, opened 2024-02-26
chore: Cleanup tests by mocking Vault class

DanielArndt:
Instead of individually patching each method, mock the entire class. This also ensures we won't "accidentally" use vault.

This PR doesn't reduce code a lot but it does:

1. Reduce the number of places things need to change
2. Make it easier to see what the entire Vault setup looks like from the perspective of the test
3. Make it easier to extract and refactor to reduce code in the future

This could be a first step in a bigger refactor. The next step would be to extract the common vault stuff into a test `setUp` for the tests that is then overridden by the tests that need "non-standard" behaviour.

# Checklist:

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -6,7 +6,7 @@ import io
 import json
 import unittest
 from typing import List
-from unittest.mock import Mock, PropertyMock, call, patch
+from unittest.mock import MagicMock, Mock, PropertyMock, call, patch
 
 import hcl  # type: ignore[import-untyped]
 import requests
@@ -23,6 +23,7 @@ from charms.tls_certificates_interface.v3.tls_certificates import (
     CertificateAvailableEvent,
     ProviderCertificate,
 )
+from charms.vault_k8s.v0.vault_client import Vault
 from charms.vault_k8s.v0.vault_tls import CA_CERTIFICATE_JUJU_SECRET_LABEL
 from ops import testing
 from ops.model import ActiveStatus, WaitingStatus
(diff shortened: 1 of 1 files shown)


tests/unit/test_charm.py
```diff
@@ -316,20 +317,14 @@ def test_given_not_leader_and_init_secret_not_set_when_configure_then_status_is_
...
 
-    @patch("charms.vault_k8s.v0.vault_client.Vault.enable_audit_device", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.is_active", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.audit_device_enabled", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.unseal", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.is_sealed", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.is_initialized", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.is_api_available", new=Mock)
-    @patch("charms.vault_k8s.v0.vault_client.Vault.is_raft_cluster_healthy", new=Mock)
+    @patch("charm.Vault", autospec=True)
```
gruyaume, line 320: I like the fact that this will prevent us from forgetting to mock any aspect of the class and end up making calls to a non existent Vault. It also makes it that each test can only focus on the aspects it cares about.

gruyaume (approved):

PR #185, opened 2024-02-26
feat: add provider side of vault-pki integration

gruyaume:
# Description

In #180 , we added the "requirer" side of the PKI integration. Here we add the "provider" side that TLS requirers can use to request TLS certificates. Vault K8s will treat those requests using the PKI secrets engine. This is accomplished using the `tls-certificates` interface which we alias to `vault-pki` here to standardise with `vault-transit` and `vault-kv`.

image

## Usage

Deploy Vault K8s

```
juju deploy vault-k8s -n 3 --config common_name=mydomain.com
```

Deploy the parent CA

```
juju deploy self-signed-certificates
```

Integrate Vault K8s with its parent CA

```
juju integrate vault-k8s:tls-certificates-pki self-signed-certificates
```

Deploy tls-certificates-requirer

```
juju deploy tls-certificates-requirer --common_name=demo.mydomain.com
```

Integrate TLS Certificates Requirer with Vault
```
juju integrate tls-certificates-requirer vault-k8s:vault-pki
```

Retrieve the certificate
```
juju run tls-certificates-requirer/leader get-certificate
```
## Note

Here we removed the default `common_name` configuration value as it made sense for 0% of the population and we want to force users to use a reasonable common name. 

## Reference

- https://developer.hashicorp.com/vault/docs/secrets/pki

# Checklist:


lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -175,6 +167,34 @@ def set_pki_intermediate_ca_certificate(self, certificate: str, mount: str) -> N
         )
         logger.info("Set the intermediate CA certificate for the PKI backend")
 
+    def sign_pki_certificate_signing_request(
+        self, mount: str, role: str, csr: str, common_name: str
+    ) -> Tuple[str, str, List[str]]:
```
ghislainbourgeois, line 172: Because this is in a lib, will this be used elsewhere? I think we might want to consider using a `dataclass` or `namedtuple` as the return type here, to make the API a bit cleaner, similar to what was done in TLS V3.

src/charm.py
```diff
@@ -418,6 +446,52 @@ def _on_tls_certificate_pki_certificate_available(self, event: CertificateAvaila
...
+            logger.error("Common name is not set in the charm config")
+            return
+        if not vault.is_pki_role_created(role=PKI_ROLE, mount=PKI_MOUNT):
+            logger.debug("PKI role not created")
+            return
+        requested_csr = csr
+        requested_common_name = get_common_name_from_csr(requested_csr)
+        if not is_requested_common_name_allowed(requested_common_name, common_name):
+            logger.warning("Requested common name [%s] is not allowed", requested_common_name)
+            return
```
ghislainbourgeois, line 480: I do not think this makes complete sense, as vault would probably have a common name like `vault.example.org`, and I would expect requirers to have names like `req1.example.org`.

Unless there is a hard requirement from Vault itself, I think I would drop this completely and use `tls-constraints` instead, or provide a way to configure policies in Vault.

ghislainbourgeois (changes requested):

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: It will be used here and it the vault machine charm.

src/charm.py
gruyaume: Ok I removed it. It means however that we may regularly make calls to the vault using the client to request for the same csr to be signed. Every time we will handle the exception and log this as a warning.

ghislainbourgeois (approved):

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -25,13 +25,26 @@
...
-LIBPATCH = 3
+LIBPATCH = 4
 
 
 logger = logging.getLogger(__name__)
 RAFT_STATE_ENDPOINT = "v1/sys/storage/raft/autopilot/state"
 
 
+class VaultClientError(Exception):
+    """Error raised when interacting with Vault."""
```
Kayra1, line 36: [hvac.exceptions.VaultError](https://hvac.readthedocs.io/en/stable/_modules/hvac/exceptions.html) is exactly this class, so there is no need for this.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -175,6 +181,41 @@ def set_pki_intermediate_ca_certificate(self, certificate: str, mount: str) -> N
...
+            )
+            logger.info("Signed a PKI certificate for %s", common_name)
+            return Certificate(
+                certificate=response["data"]["certificate"],
+                ca=response["data"]["issuing_ca"],
+                chain=response["data"]["ca_chain"],
+            )
+        except InvalidRequest as e:
+            logger.warning("Error while signing PKI certificate: %s", e)
+            raise VaultClientError("Error while signing PKI certificate: %s") from e
```
Kayra1, line 217: Do we actually want vault to crash if we fail to sign a certificate? What if the CSR is malformed? A single requirer crashing the whole charm?

src/charm.py
```diff
@@ -418,6 +446,55 @@ def _on_tls_certificate_pki_certificate_available(self, event: CertificateAvaila
...
+                mount=PKI_MOUNT,
+                role=PKI_ROLE,
+                csr=requested_csr,
+                common_name=requested_common_name,
+            )
+        except VaultClientError:
+            logger.warning(
+                "Failed to sign the PKI certificate signing request for %s", requested_common_name
+            )
+            return
```
Kayra1, line 489: I don't think having 2 excepts for 1 call makes sense. One of the reasons we have the library is so we don't have to bubble up every error hvac produces. We should just log and return None from the client when the csr can't be signed and return if not here.

Kayra1 (changes requested):
Some minor comments

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: The charm catches `VaultClientErrors` meaning this won't crash the whole charm

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -143,16 +156,9 @@ def is_intermediate_ca_set(self, mount: str, certificate: str) -> bool:
...
-        loaded_certificate = x509.load_pem_x509_certificate(intermediate_ca.encode("utf-8"))
-        existing_common_name = loaded_certificate.subject.get_attributes_for_oid(
-            x509.oid.NameOID.COMMON_NAME
-        )[0].value
-        return existing_common_name == common_name
+    def get_intermediate_ca(self, mount: str) -> str:
+        """Get the intermediate CA for the PKI backend."""
+        return self._client.secrets.pki.read_ca_certificate(mount_point=mount)
 
     def generate_pki_intermediate_ca_csr(self, mount: str, common_name: str) -> str:
```
saltiyazan, line 163: I know this was not introduced in this PR, but logs in this function seems a bit off.
Shouldn't it be something along the line: "Generated a CSR for the intermediate CA for the PKI backend"

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: Correct but it's owned by the hvac lib, it doesn't make sense to for the charm code to have to know of exceptions owned by its's libs' libs.

src/charm.py
gruyaume: I agree that we probably don't want to bubble up every error hvac produces. In this context however, we tried to do something and it failed so I felt raising was the right approach. 

I'm ok with returning None instead.

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: That makes sense. It's also possible to import the same VaultError from the client instead of hvac, but that's more of a javascript thing that doesn't happen in python.

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: Done

src/charm.py
gruyaume: Done

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: Done, we now return None here.

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: Done, we now return None in the `sign_...` method and there's no need for this error.

src/charm.py
```diff
@@ -418,6 +446,52 @@ def _on_tls_certificate_pki_certificate_available(self, event: CertificateAvaila
...
+        )
+
+    def _generate_pki_certificate_for_requirer(self, csr: str, relation_id: int):
+        """Generate a PKI certificate for a TLS requirer."""
+        if not self.unit.is_leader():
+            logger.debug("Only leader unit can handle a vault-pki certificate request")
+            return
+        if not self._is_peer_relation_created():
+            logger.debug("Peer relation not created")
+            return
```
saltiyazan, line 464: Is it really needed to check the peer relation here?

src/charm.py
```diff
@@ -418,6 +446,52 @@ def _on_tls_certificate_pki_certificate_available(self, event: CertificateAvaila
...
 
+    def _on_vault_pki_certificate_creation_request(
+        self, event: CertificateCreationRequestEvent
+    ) -> None:
+        """Handle the vault-pki certificate creation request event."""
+        self._generate_pki_certificate_for_requirer(
+            event.certificate_signing_request, event.relation_id
+        )
+
+    def _generate_pki_certificate_for_requirer(self, csr: str, relation_id: int):
```
saltiyazan, line 457: Shouldn't we check if the charm was integrated with the tls provider? 
I think this could run without having that integration in place.

src/charm.py
```diff
@@ -418,6 +446,52 @@ def _on_tls_certificate_pki_certificate_available(self, event: CertificateAvaila
         """Handle the tls-certificates-pki certificate available event."""
         self._add_ca_certificate_to_pki_secrets_engine()
 
+    def _on_vault_pki_certificate_creation_request(
+        self, event: CertificateCreationRequestEvent
+    ) -> None:
+        """Handle the vault-pki certificate creation request event."""
+        self._generate_pki_certificate_for_requirer(
```
saltiyazan, line 453: Wouldn't running this in the `_sync_vault_pki` be enough?

saltiyazan (commented):
I hope I didn't duplicate the previous comments, but here a few concerns/questions I have before approving the PR.

src/charm.py
gruyaume: Yes, we need to get info from the peer relation to initialize the Vault client

src/charm.py
gruyaume: Agreed, done

src/charm.py
gruyaume: Well that's what we do:

```python
def _sync_vault_pki(self) -> None:
        """Goes through all the vault-pki relations and sends necessary TLS certificate."""
        for relation in self.model.relations[PKI_RELATION_NAME]:
            outstanding_requests = self.vault_pki.get_outstanding_certificate_requests(
                relation_id=relation.id
            )
            for request in outstanding_requests:
                self._generate_pki_certificate_for_requirer(
                    csr=request.csr,
                    relation_id=relation.id,
                )
```

src/charm.py
saltiyazan: Do you mean `_get_initialization_secret_from_peer_relation` in `_get_initialized_vault_client`? I feel the name is a bit misleading, I know that we check for the peer relation when creating that secret and we put its id into relation data, but when getting the secret I don't think we really rely in the peer relation.
And we might get into a situation where the peer relation is created but the secret isn't. Maybe we should check for `_initialization_secret_set_in_peer_relation` instead of the peer relation itself.

saltiyazan: But in both cases, if we can't get the secret `_get_initialized_vault_client` will return `None`.

saltiyazan: It's not critical to remove it right now.

src/charm.py
saltiyazan: I meant not to have a separate handler (`_on_vault_pki_certificate_creation_request`) for `certificate_creation_request`.
Couldn't we run `_configure`?

src/charm.py
Kayra1: This is true, the peer relation is no longer required for this. But like you said it's not critical here and we can address it later. That's why I've ignored it.

src/charm.py
gruyaume: I mean, we could but it's somewhat of an overkill. Plus we'd likely want to use the same logic for both PKI and KV

saltiyazan (approved):

PR #187, opened 2024-02-27
chore: Remove repeated section in pyproject.toml

DanielArndt:
Removes a section that is repeated in `pyproject.toml`

This is an attempt to fix dependabot, which is currently complaining that `pyproject.toml` is not able to be parsed (but I can't find much more details than that).

# Checklist:


DanielArndt:
Closing because I'm not actually sure if this should be removed. Apparently this makes an array of tables in toml, which is new to me. Not sure why (or if) we need it, but since it has been around for a while it is unlikely it is the cause of the errors.

PR #209, opened 2024-03-01
chore: Rework integration tests

Kayra1:
# Description

We can reduce the chances of pods crashing and also speed up the integration tests by parallelizing charm deployments and splitting the tests into separate phases, and remove the applications that we're done with.

# Checklist:


gruyaume (approved):

gruyaume:
Sorry, I meant to approve the other PR, not this one, I haven't reviewed it properly yet

Kayra1:
No problem I won't merge until you review again.

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
+        deploy_self_signed_certificates = ops_test.model.deploy(
+            SELF_SIGNED_CERTIFICATES_APPLICATION_NAME,
+            application_name=SELF_SIGNED_CERTIFICATES_APPLICATION_NAME,
+            trust=True,
+            channel="beta",
+        )
+        deploy_tls_requirer = ops_test.model.deploy(
+            VAULT_PKI_REQUIRER_APPLICATION_NAME,
+            application_name=VAULT_PKI_REQUIRER_APPLICATION_NAME,
+            channel="edge",
```
gruyaume, line 153:
```suggestion
            channel="stable",
```

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
+        deploy_loki = ops_test.model.deploy(
+            "loki-k8s",
+            application_name=LOKI_APPLICATION_NAME,
+            trust=True,
+        )
+        deploy_self_signed_certificates = ops_test.model.deploy(
+            SELF_SIGNED_CERTIFICATES_APPLICATION_NAME,
+            application_name=SELF_SIGNED_CERTIFICATES_APPLICATION_NAME,
+            trust=True,
+            channel="beta",
```
gruyaume, line 148:
```suggestion
            channel="stable",
```

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
+    async def deploy_phase_2(self, ops_test: OpsTest):
+        deploy_prometheus = ops_test.model.deploy(
+            "prometheus-k8s",
+            application_name=PROMETHEUS_APPLICATION_NAME,
+            trust=True,
+        )
+        deploy_loki = ops_test.model.deploy(
+            "loki-k8s",
+            application_name=LOKI_APPLICATION_NAME,
+            trust=True,
```
gruyaume, line 142:
```suggestion
            trust=True,
            channel="stable",
```

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
+            channel="stable",
+        )
+        deploy_s3_integrator = ops_test.model.deploy(
+            "s3-integrator",
+            application_name=S3_INTEGRATOR_APPLICATION_NAME,
+            trust=True,
+        )
+        deploy_minio = ops_test.model.deploy(
+            "minio", application_name=MINIO_APPLICATION_NAME, trust=True, config=MINIO_CONFIG
```
gruyaume, line 84:
```suggestion
            "minio", application_name=MINIO_APPLICATION_NAME, trust=True, config=MINIO_CONFIG, channel="stable"
```

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
+        deploy_traefik = ops_test.model.deploy(
+            "traefik-k8s",
+            application_name=TRAEFIK_APPLICATION_NAME,
+            trust=True,
+            channel="stable",
+        )
+        deploy_s3_integrator = ops_test.model.deploy(
+            "s3-integrator",
+            application_name=S3_INTEGRATOR_APPLICATION_NAME,
+            trust=True,
```
gruyaume, line 81:
```suggestion
            trust=True,
            channel="stable",
```

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
-    @staticmethod
-    async def deploy_charm(ops_test: OpsTest, charm: Path) -> None:
-        """Deploys charm.
-
-        Args:
-            ops_test: Ops test Framework.
-            charm: Charm path.
-        """
+    @pytest.fixture(scope="module")
+    async def deploy_phase_1(self, ops_test: OpsTest):
```
gruyaume, line 54: We introduce the concept of a "phase" here but this is not described anywhere. What is a phase? Do we really need this concept? I understand why we want to deploy everything in parallel but we can have this without the concept of a "phase"

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
+            timeout=600,
+            wait_for_exact_units=1,
+        )
+        await ops_test.model.wait_for_idle(
+            apps=[APPLICATION_NAME],
+            status="active",
+            timeout=600,
+            wait_for_exact_units=NUM_VAULT_UNITS,
+        )
+        yield
```
gruyaume, line 120: Why is this needed?

gruyaume (changes requested):

tests/integration/test_integration.py
```diff
@@ -29,209 +31,351 @@
...
         assert ops_test.model
-        await ops_test.model.deploy(
-            "loki-k8s",
-            application_name=LOKI_APPLICATION_NAME,
-            trust=True,
+        await ops_test.model.integrate(
+            relation1=f"{APPLICATION_NAME}:vault-kv",
+            relation2=f"{VAULT_KV_REQUIRER_APPLICATION_NAME}:vault-kv",
+        )
+        await ops_test.model.wait_for_idle(
```
saltiyazan, line 240: Do we depend on the status of `VAULT_KV_REQUIRER_APPLICATION_NAME` being active? If not, I'd not check whether it becomes active, we don't have much control over that status and we are not testing it.
I know this might have been already in the tests before this PR, but I think it's worth thinking about.

saltiyazan: I know we are using a test charm for it, so if we depend on the status it's fine :)

tests/integration/test_integration.py
Kayra1: The yield is where all of the phase 1 tests run. Any line afterward is to clean up the fixture after all tests that depend on phase 1 is complete.

tests/integration/test_integration.py
Kayra1: 1) A phase of testing is simply a grouping of tests, grouped around testing the behavior of a couple of selected charms which the tests will interact with. Once a phase is over (all of the tests that call the fixture complete), those applications are removed.

2) The main reasons having phases is convenient are:
Save memory in the github runner so we don't have pods crashing and making the tests unreliable.
Split up the testing suite of charms by a logical grouping that is explicitly called out to avoid confusion.
Still parallelize as much of the building and deployment process as we can.

I'm open to renaming them or coming up with a better solution with the goal of: 
* reducing the total time it takes for the tests to complete
* managing the memory of the runner

Where should I place this description in the code?

gruyaume: I'm not convinced we need the concept of phases, let's name the methods based on what they actually do.

tests/integration/test_integration.py
gruyaume: Do we really need this? We are making the integration tests less readable through this PR. Let's make sure it's justified and we only do what really matters to address the performance issue we're observing

tests/integration/test_integration.py
gruyaume: Also, please add some docstrings to the new methods like this one.

tests/integration/test_integration.py
Kayra1: The yield keyword is explicitly called out for teardowns in the [pytest documentation](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization). It's being used here exactly as intended by pytest.

I disagree that this is less readable. Having fixtures contain their definition and teardown logic in the same function and be marked appropriately as fixtures is much clearer than having multiple different deploy functions scattered across the file, some marked as static methods when they should be fixtures, and are called in a specific order which makes the tests implicitly depend on each other.

tests/integration/test_integration.py
Kayra1: Without phases, the deployments of charms will no longer be parallel. We can't deploy every dependency together because it will crash the runner.

Other ideas are:
I can make a fixture for each of the charms and have them deploy serially, which will take much longer to deploy.
I can just create two files called integration_test_part_1.py and integration_test_part_2.py and copy and paste the duplicate code that will create.

tests/integration/test_integration.py
ghislainbourgeois: I agree with Kayra for the yield keyword. This is the proper way to do cleanup code with fixtures in Pytest.

tests/integration/test_integration.py
ghislainbourgeois: I think the phases split might not be a perfect fit here, and maybe things should be split in more fixtures, but I am not sure how to do that and keep most of the async stuff. I find mostly confusing the part where some tests depend on `phase2`, and that seems to also implicitly depend on parts of `phase1`. The cleanup of `phase1` is thus incomplete and left to the framework.

ghislainbourgeois (changes requested):
I have some concerns with the way the fixtures are split.

tests/integration/test_integration.py
```diff
@@ -29,209 +31,353 @@
...
-        """
+    async def test_given_application_is_deployed_when_scale_up_then_status_is_active(
+        self, ops_test: OpsTest, deploy_phase_1
+    ):
         assert ops_test.model
-        await ops_test.model.deploy(
-            "prometheus-k8s",
-            application_name=PROMETHEUS_APPLICATION_NAME,
-            trust=True,
+        num_units = 5
```
DanielArndt, line 207: Would it make sense for this to be `NUM_VAULT_UNITS + 2`? Incase `NUM_VAULT_UNITS` is ever changed to 5, for example?

DanielArndt (approved):
I don't mind this approach, personally. I actually find it a lot easier to read than the weird implicit relationship from before where each test relied on what fixtures were used in the previous tests. That was always frustrating to me and something I wanted to tackle at some point, so I'm glad to see some work on it.

I agree with @ghislainbourgeois that there might be a nicer way to do this with fixtures, but I also don't know off the top of my head how I would structure this. I think it's reasonable to start here, and move that direction if we want, though. I think it would be easier to take this code and move it to fixtures than it would be with the old test set-up.

Integration tests are always messy. Just my 2 cents. I'll approve to show my support, although I think it's wise to wait for some more feedback from @ghislainbourgeois and @gruyaume :laughing:

tests/integration/test_integration.py
Kayra1: I forgot I changed this for testing the memory usage, thanks for catching it.

tests/integration/test_integration.py
Kayra1: Huh, I just noticed that Traefik isn't being torn down and at the same time I mis-marked it as phase 2, and the self signed cert is also in phase 2, so the 3 bugs eliminated each other to make this thing work. That's fixed now. 

I'm not a big fan of just calling these phases but considering the alternatives I provided above, I'd still prefer going with phases for now because performance is more of a concern when testing Vault features. If you guys prefer another option above or a better paradigm, I'm open to changing it.

Kayra1: I split the tests into 3 different classes with parts, and made vault a module level fixture, and made the deployments class level. The naming is still iffy, but I think it reads much better this way and is way more pytestic or something like that.

ghislainbourgeois (approved):

gruyaume (approved):

PR #221, opened 2024-03-06
remove some unnecessary usage and wording around peer relation

Kayra1:
# Description

This PR removes some unused or unnecessary interactions with the peer relation. The usage of peer relation in KV needs special attention because it uses the relation as a dictionary of secrets, and the conversion requires changing the format of the secret too.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -34,7 +34,7 @@ from charms.vault_k8s.v0.vault_kv import NewVaultKvClientAttachedEvent, VaultKvP
 from charms.vault_k8s.v0.vault_tls import File, VaultTLSManager
 from container import Container
 from cryptography import x509
-from exceptions import PeerSecretError, VaultCertsError
+from exceptions import VaultCertsError
 from jinja2 import Environment, FileSystemLoader
 from ops.charm import (
     ActionEvent,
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -454,7 +448,7 @@ def _on_vault_pki_certificate_creation_request(
     def _generate_pki_certificate_for_requirer(self, csr: str, relation_id: int):
         """Generate a PKI certificate for a TLS requirer."""
         if not self.unit.is_leader():
-            logger.debug("Only leader unit can handle a vault-pki certificate request")
+            logger.debug("Only leader unit can handle a vault-pki certificate request")  # RFC: Why
```
gruyaume, line 451: Because we want to avoid TLS provider units competing to provide a cert, the cert is provided by the Vault app, not unit. Plus only the leader unit can write to the application relation data.

src/charm.py
```diff
@@ -1213,7 +1189,7 @@ def _bind_address(self) -> Optional[str]:
         Returns:
             str: Bind address
         """
-        peer_relation = self.model.get_relation(PEER_RELATION_NAME)
+        peer_relation = self.model.get_relation(PEER_RELATION_NAME)  # TODO: is this needed?
```
gruyaume, line 1192: The peer relation is needed to get the bind address 

```
binding = self.model.get_binding(peer_relation)
address = str(binding.network.ingress_address)
```

gruyaume (approved):

PR #234, opened 2024-03-12
feat: Vault client approle login and related improvements

Kayra1:
# Description

This change modifies vault to embrace idempotency and generalizes some functions. It also introduces some common error handling that isn't fatal, and adds the ability to login with approles.

# Checklist:


lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +130,94 @@ def is_active(self) -> bool:
...
+                options={"file_path": path},
+            )
+            logger.info("Enabled audit device %s for path %s", device_type, path)
+        except InvalidRequest:
+            logger.info("Audit device already enabled.")
 
-    def approle_login(self, role_id: str, secret_id: str) -> None:
-        """Authenticate to Vault using AppRole."""
-        self._client.auth.approle.login(role_id=role_id, secret_id=secret_id, use_token=True)
+    def enable_auth_method(self, auth_method: Literal["approle"]) -> None:
```
ghislainbourgeois, line 151: I am not convinced that having a parameter that takes a single value is a great API. Do we have other auth methods planned for support in the near future? Otherwise I would keep it simple and make this method only enable approle.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +130,94 @@ def is_active(self) -> bool:
...
-            logger.error("Error while checking Vault health status: %s", e)
-            return False
-
-    def unseal(self, unseal_keys: List[str]) -> None:
-        """Unseal Vault."""
-        for unseal_key in unseal_keys:
-            self._client.sys.submit_unseal_key(unseal_key)
-        logger.info("Vault is unsealed")
+    def enable_audit_device(
+        self, device_type: Literal["file", "syslog", "socket"], path: str
```
ghislainbourgeois, line 134: Instead of using a `Literal`, what do you think about creating an enum?

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -46,6 +75,19 @@ class Vault:
     def __init__(self, url: str, ca_cert_path: str):
         self._client = hvac.Client(url=url, verify=ca_cert_path)
 
+    def authenticate(self, auth_details: Union[Token, AppRole]) -> None:
```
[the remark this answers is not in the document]
Kayra1: I've managed to avoid using protocols up until this point because I've found them esoteric, but I'm not opposed to that if we plan to support more auth methods in the long run.

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: I liked the readability and the consistency with other methods of `enable_auth_method('approle')` more than `enable_approle_auth_method()`, but I'm not opposed to changing it. I don't think either charm will ever use anything other than approles, although there may be a case where a user may want to enable it through the charm.
Although that's very far away in the future.

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: I considered using an enum, but felt that it was too wordy and the string conversions are a bit implicit, and the Literal type had the ergonomics to use this function well enough. I do not have a strong opinion either way.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +130,94 @@ def is_active(self) -> bool:
...
+            policy_path: The path of the file where the policy is defined, ending with .hcl
+            mount: Formats the policy string with the given mount path to create dynamic policies
+        """
+        with open(policy_path, "r") as f:
+            policy = f.read()
+        self._client.sys.create_or_update_policy(
+            name=policy_name,
+            policy=policy if not mount else policy.format(mount=mount),
+        )
+        logger.debug(f"Created or updated charm policy: {policy_name}")
```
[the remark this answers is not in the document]
Kayra1: Fixed.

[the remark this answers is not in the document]
Kayra1: fixed

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: Changed.

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: Changed.

Kayra1: Funnily enough, the type hint for the enum is `(property) value: Literal['file', 'syslog', 'socket'`

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +130,94 @@ def is_active(self) -> bool:
...
+
+    def read_role_secret(self, name: str, id: str) -> dict:
+        """Get definition of a secret tied to an AppRole."""
+        response = self._client.auth.approle.read_secret_id(name, id)
+        return response["data"]
 
-    def is_secret_engine_enabled(self, path: str) -> bool:
-        """Check if a PKI mount is enabled."""
-        return path + "/" in self._client.sys.list_mounted_secrets_engines()
+    def enable_secrets_engine(self, backend_type: Literal["kv-v2", "pki"], path: str) -> None:
```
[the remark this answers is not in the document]
Kayra1: Same comments as above, but I changed it.

ghislainbourgeois (approved):

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
Kayra1: I did end up adding the protocol, not sure if you've seen it.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +160,88 @@ def is_active(self) -> bool:
...
-    def approle_login(self, role_id: str, secret_id: str) -> None:
-        """Authenticate to Vault using AppRole."""
-        self._client.auth.approle.login(role_id=role_id, secret_id=secret_id, use_token=True)
+    def enable_approle_auth_method(self) -> None:
+        """Enable approle auth method."""
+        try:
+            self._client.sys.enable_auth_method("approle")
+            logger.info("Enabled approle auth method.")
+        except InvalidRequest:
+            logger.info("Approle already enabled.")
```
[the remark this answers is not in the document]
Kayra1: Good question. The exceptions in `hvac` are mostly a mapping of the [HTTP status codes](https://developer.hashicorp.com/vault/api-docs#400) that vault returns. The confidence comes from knowing that we filter the wrong parameters before we ever make the call to `hvac` in the first place. If a regression happens where this exception is being raised because of a different reason, the solution would be to filter that out in python before we make the call to `hvac` rather than having to read and filter through the InvalidRequest exception.

The assumptions here could change in the future, but we can handle that in the future when this method does do what you say it might.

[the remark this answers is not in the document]
saltiyazan: It seems to me that we want to rely on these exceptions to validate that approle or audit devices were actually enabled and that can lead to false positives. This is my main concern.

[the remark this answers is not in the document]
Kayra1: I thought about this a bit more, and decided that we may as well do it now. changed.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -24,28 +26,94 @@
...
+
+    role_id: str
+    secret_id: str
+
+    def login(self, client: hvac.Client):
+        """Authenticate a vault client with approle details."""
+        client.auth.approle.login(role_id=self.role_id, secret_id=self.secret_id, use_token=True)
+
+
+class AuthMethod(Protocol):
```
DanielArndt, line 65: I don't really understand the benefit of using protocols here. I know @ghislainbourgeois recommended them but it's not clear to me what advantage this has over something like an ABC which makes the descendants explicit?

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +160,100 @@ def is_active(self) -> bool:
...
-        """Unseal Vault."""
-        for unseal_key in unseal_keys:
-            self._client.sys.submit_unseal_key(unseal_key)
-        logger.info("Vault is unsealed")
-
-    def set_token(self, token: str) -> None:
-        """Set the Vault token for authentication."""
-        self._client.token = token
+    def enable_audit_device(self, device_type: AuditDeviceType, path: str) -> None:
+        """Enable a new audit device at the supplied path.
```
DanielArndt, line 164: IMO it's worth noting the idempotentcy in the docstring, i.e 

```suggestion 
        """Enable a new audit device at the supplied path if it isn't already enabled.
```

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +160,100 @@ def is_active(self) -> bool:
...
+            logger.info("Enabled audit device %s for path %s", device_type.value, path)
+        except InvalidRequest as e:
+            errors = e.json.get("errors")
+            if len(errors) == 1 and errors[0].startswith("path is already in use"):
+                logger.info("Audit device already enabled.")
+            else:
+                raise e
+
+    def enable_approle_auth_method(self) -> None:
+        """Enable approle auth method."""
```
DanielArndt, line 184:
```suggestion
        """Enable approle auth method if it isn't already enabled."""
```

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +160,100 @@ def is_active(self) -> bool:
...
+            else:
+                raise e
 
-    def configure_charm_access_policy(self, name: str) -> None:
-        """Create/update a policy within vault for the charm to access vault."""
-        with open("src/templates/charm_policy.hcl", "r") as f:
-            charm_policy = f.read()
-        self._client.sys.create_or_update_policy(name=name, policy=charm_policy)
-        logger.debug(f"Created charm policy: {name}")
+    def configure_policy(
```
DanielArndt, line 195: nit: Not a huge fan of these names (`configure_` when they're really creating/updating, but I realize that's a holdover. Feel free to ignore.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -82,80 +160,100 @@ def is_active(self) -> bool:
...
+        """Create/update a policy within vault.
 
-    def remove_raft_node(self, node_id: str) -> None:
-        """Remove raft peer."""
-        self._client.sys.remove_raft_node(server_id=node_id)
-        logger.info("Removed raft node %s", node_id)
+        Args:
+            policy_name: Name of the policy to create
+            policy_path: The path of the file where the policy is defined, ending with .hcl
+            mount: Formats the policy string with the given mount path to create dynamic policies
```
DanielArndt, line 203: Bonus points for not including the type here when it's already type hinted :laughing:

DanielArndt (approved):

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: A protocol is kinda like an interface than an ABC and Python internally uses them a lot, like `Iterable`. It's all just fancy type hints at the end of the day so I don't really mind as long as it's readable.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
saltiyazan: Wouldn't it be a more robust approach to actually check if these are enabled instead of relying on the exception?

PR #261, opened 2024-03-25
chore: use update status event to sync kv requests

gruyaume:
# Description

Here we update the `vault_kv` library to make it possible for the charm to get kv requests at any moments using `getters`. The charm uses those getters to sync kv requests on update status events, just like we do for tls certificate requests.

## Rationale
1. Move all relation data logic to the kv library. You won't see any `relation.data.get()` in charm code anymore. It all lives in the `vault_kv` lib, where it belongs.
2. Standardise resource provider approach between tls certificates and kv secrets, both offer pull based mathods allowing to get the outstanding requests at any moment.
3. Get rid of `event.defer()` for kv requests. 
4. This is a major change for the provider side of the lib for which only vault-k8s uses. The intent was to make this change before using it in the machine version of the charm to avoid back and forths. 

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_kv.py
+++ b/lib/charms/vault_k8s/v0/vault_kv.py
@@ -201,7 +201,7 @@ class KVRequest:
     nonce: str
 
 
-def is_requirer_data_valid(app_data: dict, unit_data: dict) -> bool:
+def is_requirer_data_valid(app_data: Mapping[str, str], unit_data: Mapping[str, str]) -> bool:
     """Return whether the requirer data is valid."""
     try:
         RequirerSchema(
@@ -214,7 +214,7 @@ def is_requirer_data_valid(app_data: dict, unit_data: dict) -> bool:
         return False
 
 
-def is_provider_data_valid(data: dict) -> bool:
+def is_provider_data_valid(data: Mapping[str, str]) -> bool:
     """Return whether the provider data is valid."""
     try:
         ProviderSchema(app=VaultKvProviderSchema(**data))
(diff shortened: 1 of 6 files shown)


lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -351,6 +371,53 @@ def get_credentials(self, relation: ops.Relation) -> dict:
...
+                if relation.id == relation_id
+            ]
+            if relation_id is not None
+            else self.model.relations.get(self.relation_name, [])
+        )
+        for relation in relations:
+            app_data = dict(relation.data[relation.app])
+            for unit in relation.units:
+                unit_data = dict(relation.data[unit])
+                if not is_requirer_data_valid(app_data=app_data, unit_data=unit_data):
```
ghislainbourgeois, line 399: I saw those `dict` creation from the relation data and noticed that the typing for `is_requirer_data_valid` specifies `dict`. We might be able to change those type hints to `Mapping` instead, and we could save creating new dictionaries from the relation data and just use the returned mapping directly. This is pretty minor however, so I will not block the PR for this.

src/charm.py
```diff
@@ -401,15 +372,62 @@ def _add_ca_certificate_to_pki_secrets_engine(self) -> None:
...
+        ca_certificate = self.tls.pull_tls_file_from_workload(File.CA)
+        if not ca_certificate:
+            logger.debug("Vault CA certificate not available")
+            return
+        vault = self._get_initialized_vault_client()
+        if not vault:
+            logger.debug("Failed to get initialized Vault")
+            return
+        vault.enable_approle_auth_method()
+        mount = "charm-" + app_name + "-" + mount_suffix
```
ghislainbourgeois, line 426: Pretty minor again, but this would be more efficient by not creating 3 different strings.
```suggestion
        mount = f"charm-{app_name}-{mount_suffix}"
```

ghislainbourgeois (approved):
Approved with some minor suggestions.

lib/charms/vault_k8s/v0/vault_kv.py
gruyaume: Done

PR #265, opened 2024-03-26
build: Static check all the code

DanielArndt:
Includes tests and libs owned by this charm.

# Checklist:


lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -178,6 +178,8 @@ def enable_audit_device(self, device_type: AuditDeviceType, path: str) -> None:
             )
             logger.info("Enabled audit device %s for path %s", device_type.value, path)
         except InvalidRequest as e:
+            if not e.json:
+                raise VaultClientError(e) from e
```
Kayra1, line 182: didn't know you could do `from` like this. Why wasn't any of this caught by tox before?

DanielArndt: The `from` style has nothing to do with tox. It didn't catch that, that is a change I made last week.

Unrelated, tox wasn't checking any of these files because calling `pyright` without any parameters just checks `src/` for some reason.

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -367,7 +369,7 @@ def remove_unit_credentials(self, relation: ops.Relation, nonce: Union[str, Iter
 
         relation.data[self.charm.app]["credentials"] = json.dumps(credentials, sort_keys=True)
 
-    def get_credentials(self, relation: ops.Relation) -> Mapping[str, str]:
+    def get_credentials(self, relation: ops.Relation) -> Any:
```
saltiyazan, line 372: what was the issue with the Mapping?

saltiyazan (approved):
Other than a minor comment it looks good

saltiyazan (dismissed):
I approved and then noticed that static checks are not passing.

DanielArndt:
Oh this is a fun one. I went on a little journey here investigating why this broke.

The reason this is now failing is because [someone recently published type stubs for hvac to typeshed](https://github.com/python/typeshed/pull/11591).

The latest version was of the typeshed stubs was [pulled into pyright moments later](https://github.com/microsoft/pyright/commit/6ef2ca1b27503fa110c56c5cfd21872ed5d91538).

Unfortunately, these new type stubs identify the `json` attribute on `VaultError` simply as `object`. It should likely be `dict | None`, but I'm not positive about that. I'll need to defer making a pull request to `typeshed` until I've had some time to read more about the process.

In the meantime, I will add an `isinstance` check to make sure the `json` attribute is a dict, but this wouldn't be necessary if the type stubs were correct.

lib/charms/vault_k8s/v0/vault_kv.py
DanielArndt: `json.loads()` won't return a `Mapping`, so it was complaining. I believe a `Mapping` is immutable.

`json.loads()` returns `Any`, although for an object like this it will always convert it to a `dict`. We should be safe to assume a `dict`, since pydantic will enforce that on the way in. 

I'll change this to `dict`.

Kayra1 (approved):
Did this actually catch anything? It looks like it just added a bunch of asserts and isinstances to things that are never going to fail, ignore type comments and pyright specific bs and bug comments.

DanielArndt:
> Did this actually catch anything? It looks like it just added a bunch of asserts and isinstances to things that are never going to fail, ignore type comments and pyright specific bs and bug comments.

Yes, I think it did. Two things I think were important:

1. The `Mapping` type annotation was wrong, so if the user had properly interpreted it, they would assume they had an immutable value.
2. If a `VaultError` didn't have a `json` attribute (which is possible according to the contructor), then `vault_client` would have raised an error while handling an error, and wouldn't have properly re-branded it to a `VaultClientError` -- which means our charm would crash.

PR #322, opened 2024-04-22
fix: Moves boto3 related dependencies to requirements.in

saltiyazan:
# Description

Moves boto3 related dependencies to requirements.in
boto3-stubs is a transitive dependency of types-boto3 that is used in the test-requirements, therefore to avoid any conflicts we keep all boto3 related dependencies in one place.

# Checklist:

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -12,4 +12,3 @@ requests
 jsonschema
 cryptography
 boto3
-boto3-stubs[s3]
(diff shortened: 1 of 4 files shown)


Kayra1 (approved):

DanielArndt (changes requested):
`boto3-stubs` and `types-boto3` are the same thing. We should just remove `types-boto3` which is a proxy package.

But it should be in `test-requirements.in`, not `requirements.in`.

DanielArndt (approved):

PR #332, opened 2024-04-23
fix: check for running Vault service correctly

DanielArndt:
Before we just retrieved the service object, so the function would always return true if the service existed.

This was masked by the way we delegate methods to the underlying `OpsContainer`, which is probably worth refactoring, but not necessary for this fix.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -605,7 +605,9 @@ class VaultCharm(CharmBase):
         )
         if not content_uploaded:
             event.fail(message="Failed to upload backup to S3 bucket.")
-            logger.error("Failed to run create-backup action - Failed to upload backup to S3 bucket.")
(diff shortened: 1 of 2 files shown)


gruyaume (approved):
Looks good, shouldn't most of these lint issues be caught by ruff?

DanielArndt:
> Looks good, shouldn't most of these lint issues be caught by ruff?

Yeah I was a little confused by that as well. My first guess would be that people are running different versions of ruff which might format things slightly differently (but without any explicit rules that would cause a failure). But I didn't check. I just made sure that this PR was done with the version of ruff provided in tox. 

I'll take a deeper look and see if I can figure out what is causing this, because I've seen it on a few PRs now.

PR #342, opened 2024-04-25
feat: add audit logs to dashboard

Kayra1:
# Description

This change modifies the grafana dashboard configuration to add audit logs from Vault.

Screenshot 2024-04-25 at 13 18 00

# Checklist:

### DIFF
--- a/src/grafana_dashboards/vault.json
+++ b/src/grafana_dashboards/vault.json
@@ -20,7 +20,6 @@
     "fiscalYearStartMonth": 0,
     "gnetId": 12904,
     "graphTooltip": 1,
-    "id": 12,
     "links": [],
     "liveNow": false,
     "panels": [
@@ -105,7 +104,7 @@
                         "uid": "${prometheusds}"
                     },
                     "editorMode": "builder",
-                    "expr": "max(vault_autopilot_healthy{juju_application=~\"$juju_application\", juju_model=~\"$juju_model\", juju_model_uuid=~\"$juju_model_uuid\", juju_unit=~\"$juju_unit\"})",
+                    "expr": "max(vault_autopilot_healthy{juju_application=~\"$juju_application\",juju_model=~\"$juju_model\",juju_model_uuid=~\"$juju_model_uuid\",juju_unit=~\"$juju_unit\"})",
                     "hide": false,
                     "legendFormat": "__auto",
                     "range": true,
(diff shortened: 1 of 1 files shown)


src/grafana_dashboards/vault.json
```diff
@@ -858,6 +857,95 @@
...
+                "wrapLogMessage": false
+            },
+            "targets": [
+                {
+                    "datasource": {
+                        "type": "loki",
+                        "uid": "PCFE195BA03E127CE"
+                    },
+                    "editorMode": "code",
+                    "expr": "{charm=\"vault-k8s\"} | json | type = `response` or type = `request`",
```
gruyaume, line 905: I don't think we need to be explicit about the charm name here. We can use `$juju_application` to represent it.

src/grafana_dashboards/vault.json
```diff
@@ -858,6 +857,95 @@
...
+                "showLabels": false,
+                "showTime": false,
+                "sortOrder": "Descending",
+                "wrapLogMessage": false
+            },
+            "targets": [
+                {
+                    "datasource": {
+                        "type": "loki",
+                        "uid": "PCFE195BA03E127CE"
```
gruyaume, line 939: Similar comment, we shouldn't be explicit about the loki uid. Please look at other panels here.

gruyaume (changes requested):

src/grafana_dashboards/vault.json
Kayra1: Updated

src/grafana_dashboards/vault.json
Kayra1: If I use any combination of the `$juju_application`, `$juju_model`, `$juju_model_uuid` or `$juju_unit`, I get an error saying Loki does not support regexes that include the empty string. In fact, we can see that it is included in the other queries:
`sum(vault_secret_kv_count{juju_application=~".*",juju_model=~".*",juju_model_uuid=~".*",juju_unit=~".*"})`

I can either create a new variable that doesn't include the empty string, or change the regexes to be ".+" instead of ".*"

Kayra1: also, this regex will match all applications. If there was a different application that provided logs to this loki datasource with a field `type = response or type = request`, it would also match and show up here.

gruyaume: This sounds like a bug, you may want to create an issue in the Grafana charm
- https://github.com/canonical/grafana-k8s-operator

Kayra1: No, it's meant to be this way 
```
parse error : queries require at least one regexp or equality matcher that does not have an empty-compatible value. For instance, app=~".*" does not meet this requirement, but app=~".+" will
```

https://github.com/grafana/loki/issues/7816

gruyaume: I'm not saying loki should accept an empty string. My point is that the $juju_application variable should exist for the loki data source (?)

Kayra1: It does exist, it's a query variable with the regex ".*" for the prometheus source. This regex is incompatible with loki. Using ".+" instead matches any non-empty string, which doesn't break the rest of the panels and also makes this variable work properly with loki.

Kayra1: I've updated this to remove the regexes from the variables. Everything is working as expected and we still have variables everywhere, let me know what you think.

gruyaume (approved):

PR #346, opened 2024-04-25
feat: Add vault-autounseal library

DanielArndt:
Adds the vault-autounseal library according to TE047 spec.

This library abstracts all the autounseal specific functionality that doesn't belong directly in a charm.

# Checklist:


lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume, line 1: Should we bring this PR back to life (and update its name to be `vault-autounseal`)?
- https://github.com/canonical/charm-relation-interfaces/pull/141

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+            self.charm.on[relation_name].relation_broken, self._on_relation_broken
+        )
+
+    def _on_relation_created(self, event: ops.RelationCreatedEvent) -> None:
+        self.on.vault_autounseal_initialize.emit(relation=event.relation)
+
+    def _on_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
+        self.on.vault_autounseal_destroy.emit(relation=event.relation)
+
+    def set_vault_url(self, relation: ops.Relation, address: str) -> None:
```
gruyaume, line 222: Is there value in having a bunch of independent sets instead of consolidating them together?

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+        relation.data[self.charm.app].update({"credentials_secret_id": secret_id})
+
+    def set_ca_certificate(self, relation: ops.Relation, ca_certificate: str) -> None:
+        """Set the ca_certificate on the relation.
+
+        Args:
+            relation: The relation for which to set the ca_certificate.
+            ca_certificate: The CA certificate to use when validating the Vault server's certificate.
+        """
+        if not self.charm.unit.is_leader():
```
gruyaume, line 255: Why do we have different validations here vs. in `set_credentials_secret_id` for example?

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+        self.relation = relation_name
+
+        self.framework.observe(charm.on[relation_name].relation_changed, self._on_relation_changed)
+
+    def _on_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        if event.app is None:
+            logger.warning("No remote application yet")
+            return
+        if is_provider_data_valid(event.relation.data[event.app]):
+            self.on.vault_autounseal_details_ready.emit(event.relation)
```
gruyaume, line 301: Why pass the whole relation instead of the actual content that is needed? What would the requirer do with the relation object?

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def get_vault_url(self, relation: ops.Relation) -> Optional[str]:
```
gruyaume, line 303: Why not take fetch the relation directly instead of asking it as a parameter? We already know the relation name from the `__init__`. This applies at a couple of places.

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+            relation: The relation from which to get the credentials secret id.
+
+        Returns:
+            The credentials secret id if it exists, None otherwise.
+        """
+        if relation.app is None:
+            return None
+        return relation.data[relation.app].get("credentials_secret_id")
+
+    def get_credentials(self, credentials_secret_id: str) -> tuple[str | None, str | None]:
```
gruyaume, line 330: Does this need to be public? We expect the requirer to interface with `get_credentials_from_relation` right?

tests/unit/lib/charms/vault_k8s/v0/test_vault_autounseal.py
```diff
@@ -0,0 +1,229 @@
...
+        self.harness = testing.Harness(
+            VaultAutounsealProviderCharm, meta=VaultAutounsealProviderCharm.metadata_yaml
+        )
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+
+    def setup_relation(self, leader: bool = True) -> tuple:
+        """Set up a relation between the charm and a remote app with 1 unit."""
+        remote_app = "vault-autounseal-requires"
+        remote_unit = remote_app + "/0"
```
gruyaume, line 88:
```suggestion
        remote_unit =  f"{remote_app}/0"
```

gruyaume: Use of f-string for consistency with other parts of the code.

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: Yes, planned on updating that once agreement once made in this PR.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: I don't have a strong opinion here, I just followed the same pattern as vault_kv.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: Once again, mostly just modelled off how `vault_kv` was doing things.

I admittedly don't have a great intuition on what the edge cases are for most of these, so I tried to conform.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: Well, we would need at least the ID from the relation to properly configure autounseal. We could do that and also return the databag contents directly, but we actually won't use this event other than to fire `_configure()` anyway. 

If we wanted it implement a standalone handler in the future (for some reason, if it becomes needed), I figured we could get everything we needed from the relation (since there are already getters that take the relation, which we need for `_configure`)

Oh, and you guessed it, I also copied it from vault_kv :laughing: .

Happy to change if you think it would look cleaner one way or the other.

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume: I don't know that `vault_kv` is the best of our Python library, I wouldn't take too much from it :laughing:

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: We could. I assumed there would be some performance benefit to not retrieving this every time. Once again following precedence in vault_kv.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: I don't think it really matters. Does it need to be private? I don't see any point in restricting people not to use this (and also, vault_kv of course).

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+
+The requirer charm is the charm that wishes to autounseal a Vault instance via
+the Vault transit backend.
+
+### Integration
+
+You can integrate both charms by running:
+
+```bash
+juju integrate <vault a>:vault-autounseal-provides <vault b>:vault-autounseal-requires
```
gruyaume, line 41: Here we document the interface name which is dependent on how we name those in metadata.yaml. If we want to force those to be this value, we may want to add some content in the docs here that relates to metadata.yaml.

tests/unit/lib/charms/vault_k8s/v0/test_vault_autounseal.py
DanielArndt: Carry over from vault_kv :laughing:  -- I'll change.

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+        if not self.charm.unit.is_leader():
+            return
+        if not relation:
+            logger.warning("Relation is None")
+            return
+        if not relation.active:
+            logger.warning("Relation is not active")
+            return
+        relation.data[self.charm.app]["ca_certificate"] = ca_certificate
+
```
gruyaume, line 264: We should probably have a `get_outstanding_transit_requests` or something along those lines. For instance if 3 relations gets created before Vault A is up, no event will inform it that it has those outstanding requests to satisfy.

On configure, Vault A will likely have to call this method, and go through those outstanding requests and satisfy those.

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+    """Events raised by the vault-autounseal relation on the provider side."""
+
+    vault_autounseal_initialize = ops.EventSource(VaultAutounsealInitialize)
+    vault_autounseal_destroy = ops.EventSource(VaultAutounsealDestroy)
+
+
+class VaultAutounsealRequireEvents(ops.ObjectEvents):
+    """Events raised by the vault-autounseal relation on the requirer side."""
+
+    vault_autounseal_details_ready = ops.EventSource(VaultAutounsealDetailsReadyEvent)
```
Kayra1, line 196: Do we expect requirers to just use `relation_broken`?

lib/charms/vault_k8s/v0/vault_autounseal.py
Kayra1: Wouldn't they create 3 `relation_created` events which create 3 `vault_autounseal_initialize` events in the charm?

Not that it changes the fact that if one of these fails there is no path to recovery (without defer), which there should be.

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,371 @@
...
+    on: VaultAutounsealRequireEvents = VaultAutounsealRequireEvents()  # type: ignore
+
+    def __init__(self, charm: ops.CharmBase, relation_name: str):
+        super().__init__(charm, relation_name)
+        self.relation = relation_name
+
+        self.framework.observe(charm.on[relation_name].relation_changed, self._on_relation_changed)
+
+    def _on_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        if event.app is None:
```
Kayra1, line 297: is this even possible?

lib/charms/vault_k8s/v0/vault_autounseal.py
Kayra1: I like passing it in as a parameter because we don't have to repeat `if not` statements and rely on the static type checker to warn if it's `type | None` if used improperly.

Kayra1 (commented):
Just some questions and comments

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/0651467cd3a75065ba0f04391194ceef96844526

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/0651467cd3a75065ba0f04391194ceef96844526

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/0651467cd3a75065ba0f04391194ceef96844526

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/0651467cd3a75065ba0f04391194ceef96844526

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: no longer relevant in https://github.com/canonical/vault-k8s-operator/pull/346/commits/0651467cd3a75065ba0f04391194ceef96844526

tests/unit/lib/charms/vault_k8s/v0/test_vault_autounseal.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/0651467cd3a75065ba0f04391194ceef96844526

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: That's a good point. I it doesn't really matter what they call it, I suppose. If the intention is unclear, do you have any suggestions for a better way to document this? Should I just remove the interface name entirely like we do with other libs?

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: > Wouldn't they create 3 relation_created events which create 3 vault_autounseal_initialize events in the charm?

That's what I thought would happen. I thought these events were promised to fire by the Juju framework. 

> Not that it changes the fact that if one of these fails there is no path to recovery (without defer), which there should be.

Using defer was my plan, here. It seems like a case where it makes sense. 

I really don't like the implication of trying to catch _every_ scenario of un-fired events in `_configure`.. but if that's the direction we're going, I'll reluctantly add it :laughing:

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: That's a good question. I admittedly overlooked this because it's effectively handled by `_configure` anyway the next time an event fires.

I'll add an event for this. Makes sense.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: No :stuck_out_tongue:

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume: If it makes sense to use defer here, it makes sense to use it in every relation that Vault provides. If we want to change the direction for defers we should be intentional about it, spec it,  and convince the team and me that it is a better idea than the current pattern. 

For the time being I would encourage following the same pattern we have for other interfaces.

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume: > Well, we would need at least the ID from the relation to properly configure autounseal.

Why does the requirer need the ID? It needs the vault url and credentials right? Why not pass those instead? I think we will likely limit this relation to `1` from the requirer side removing any ambiguity. This way we should be able to run run `self.vault_autounseal.get_vault_url()`?

In any case, if we don't plan on using the event handler at all, I don't think it makes sense to keep it.

DanielArndt: The path to the transit key is based on the relation id, but I suppose we could (and probably should) just add a method to get that.

To properly configure it will need

- vault url
- approle id
- approle secret id
- ca cert
- key  name (based on relation id)
- mount path (constant)

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/346/commits/c9675c04e4090f53cf9f9fd62a6b9463875ca6a4

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/346/commits/c9675c04e4090f53cf9f9fd62a6b9463875ca6a4

Except the part about adding a way to get the key path instead of inferring it, that will come in a later commit.

DanielArndt: Transit key path method added in https://github.com/canonical/vault-k8s-operator/pull/346/commits/14fd1ad51eb98cd6682bdc5af57dfe29e709cec1

lib/charms/vault_k8s/v0/vault_autounseal.py
Kayra1: I'm open to talking about it again.

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,457 @@
+#!/usr/bin/env python3
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Library for the vault-autounseal relation.
```
saltiyazan, line 5: Could you please provide an example on how the requirer and the provider charms would implement this library?

lib/charms/vault_k8s/v0/vault_autounseal.py
saltiyazan: Why did [7e9e247](https://github.com/canonical/vault-k8s-operator/pull/346/commits/7e9e247c566b543488e6c7676467f709e3652ecb) remove the changes introduced in [c9675c0](https://github.com/canonical/vault-k8s-operator/commit/c9675c04e4090f53cf9f9fd62a6b9463875ca6a4) of passing the information to the event?

DanielArndt: Because I did it late at night :laughing: 

Oops, I must I have done something wrong when switching between branches. Fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/c23518ce430bfb6d5566c1e20e0d1acc4247c1a8

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume: We can add a section in the requirer side of the lib with content similar to this: 

```
Add the following to `metadata.yaml`:
    requires:
      vault-autounseal-requires:
        interface: vault-autounseal
```

And same thing for the provider.

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,469 @@
...
+
+    def _credentials_issued_for_request(self, relation_id: Optional[int]) -> bool:
+        """Return whether credentials have been issued for the request."""
+        relation = self.model.get_relation(self.relation_name, relation_id)
+        if not relation:
+            return False
+        credentials = self.get_credentials(relation)
+        return all(credentials)
+
+    def get_credentials(self, relation) -> tuple[str | None, str | None]:
```
gruyaume, line 328:
```suggestion
    def get_credentials(self, relation: ops.Relation) -> tuple[str | None, str | None]:
```

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,469 @@
...
+        if is_provider_data_valid(data):
+            role_id, secret_id = self.get_credentials()
+            self.on.vault_autounseal_details_ready.emit(
+                data["address"], role_id, secret_id, data["ca_certificate"]
+            )
+
+    def _on_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
+        self.on.vault_autounseal_provider_removed.emit()
+
+    def get_vault_url(self) -> Optional[str]:
```
gruyaume, line 406: If we need all the relation content at the same time from the tls requirer point of view, should we combine those `gets` into a single `self.vault_autounseal.get_information()`? We may also want a `self.vault_autounseal.is_information_available()`.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/346/commits/f46a5faef9cf31c600b380bf8e809225e9cceb3e

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: Is this what you were thinking?

https://github.com/canonical/vault-k8s-operator/pull/346/commits/c4f31eaaf52fc32123a4ceeb2f377006529a6a71

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,488 @@
...
+The requirer charm is the charm that wishes to autounseal a Vault instance via
+the Vault transit backend.
+
+Add the following to `metadata.yaml`:
+
+```yaml
+requires:
+    vault-autounseal-requires:
+    interface: vault-autounseal
+    limit: 1
```
gruyaume, line 50:
```suggestion
requires:
  vault-autounseal-requires:
    interface: vault-autounseal
    limit: 1
```

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,488 @@
...
+
+The provider charm is the charm that provides a Vault instance that can be
+used to autounseal other Vault instances via the Vault transit backend.
+
+Add the following to `metadata.yaml`:
+
+```yaml
+provides:
+    vault-autounseal-requires:
+    interface: vault-autounseal
```
gruyaume, line 36:
```suggestion
provides:
  vault-autounseal-provides:
    interface: vault-autounseal
```

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume: Yes, I just added a suggestion to fix the formatting + a typo

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: Oops, guess de-indent didn't work so well there...

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: I'm more than happy to do this, but maybe this documentation is already more than we need. I kind of lost sight of this while addressing other comments as well.

Since this library is only something we will use internally (or, at least that's the intention for now), it is more akin to `vault_client` and `vault_tls` which have very minimal documentation.

What do you think?

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/346/commits/cc18132656fddc96d0e43cd286454ac01c149ee9

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/346/commits/cc18132656fddc96d0e43cd286454ac01c149ee9

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,488 @@
...
+        outstanding_requests: List[Relation] = []
+        requirer_requests = self.get_requirer_requests(relation_id=relation_id)
+        for relation in requirer_requests:
+            if not self._credentials_issued_for_request(relation_id=relation.id):
+                outstanding_requests.append(relation)
+        return outstanding_requests
+
+    def get_requirer_requests(self, relation_id: Optional[int] = None) -> List[Relation]:
+        """Get all requests for the relation."""
+        pending: List[Relation] = []
```
gruyaume, line 317: A nit but `pending` suggests that the request is pending but it could be already fulfilled (i.e. not pending).

DanielArndt: nits are my favourite

yeah I think I named this when  1, it had more logic, and 2, I didn't quite follow the code I was modelling it after yet.

This code can be simplified, so I've done that instead.

DanielArndt: No longer relevant in https://github.com/canonical/vault-k8s-operator/pull/346/commits/aabbbfdc8b866409c01e06f79c6b288a1aef17b1

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: I think I should be covered as the only place I'll want to check this, I'll need the details anyway if they exist, so I'll probably opt for something along the lines of 

```
details = self.requirer.get_details()
if not all(details):
     ...
```

Unless you think that's a bad approach?

That might change a little bit as I'm just considering if I should use a dataclass or namedtuple or something to clean up the return type of `get_details()`

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -0,0 +1,453 @@
...
+        super().__init__(charm, relation_name)
+        self.relation_name = relation_name
+
+        self.framework.observe(charm.on[relation_name].relation_changed, self._on_relation_changed)
+        self.framework.observe(charm.on[relation_name].relation_broken, self._on_relation_broken)
+
+    def _on_relation_changed(self, event: ops.RelationChangedEvent) -> None:
+        data = event.relation.data[event.app]
+        if is_provider_data_valid(data):
+            self.on.vault_autounseal_details_ready.emit(*self.get_details())
```
saltiyazan, line 411: wouldn't `*self.get_details()` make it a bit flakey?
For example if we switch the order of address and ca_certificate in the tuple returned from `get_details` we won't get any error but the user would get the ca_certificate at event.address and the address at event.ca_certificate? or is that not true?

lib/charms/vault_k8s/v0/vault_autounseal.py
saltiyazan: I find these examples useful to understand the design of the lib as I see an example of how it is being used.
But I can get the same information from unit tests, so no need to add it here too.

saltiyazan (approved):
Looks good overall, I left one observation.

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: That is true. 

That's probably a good argument for following up on namedtuple or a dataclass as mentioned in [this comment](https://github.com/canonical/vault-k8s-operator/pull/346#discussion_r1583634684). I'll change this up.

DanielArndt: @saltiyazan What are your thoughts on this? https://github.com/canonical/vault-k8s-operator/pull/346/commits/3622bef0686a00cb7270aa6785136343a67eded1

saltiyazan: Seems more difficult to mess things up and not notice it now

PR #361, opened 2024-05-06
fix: Emits connected event on a refresh event

### LINKED ISSUE #337: Unauthorized through CIDR restrictions on the secret ID
### Bug Description

See discussion [here](https://matrix.to/#/!yAkGlrYcBFYzYRvOlQ:ubuntu.com/$iSh81WI9kAhzAnXzDoPjm4_yGu61Qcb8p7GnoC8Zeo8?via=ubuntu.com&via=matrix.org&via=mozilla.org)
```

### To Reproduce

See discussion [here](https://matrix.to/#/!yAkGlrYcBFYzYRvOlQ:ubuntu.com/$iSh81WI9kAhzAnXzDoPjm4_yGu61Qcb8p7GnoC8Zeo8?via=ubuntu.com&via=matrix.org&via=mozilla.org)

### Environment

microk8s

### Relevant log output

```shell
https://pastebin.canonical.com/p/Xk9qjgfCWG/
```

### Additional context

_No response_

saltiyazan:
# Description

The aim is to allow the requirer charm to provide a refresh event that will trigger the `connected` event where the requirer can re-request the credentials with new IPs.

Fixes #337 

# Checklist:


lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -519,15 +533,19 @@ def __init__(
         charm: ops.CharmBase,
         relation_name: str,
         mount_suffix: str,
+        refresh_events: Optional[List] = None,
     ) -> None:
         super().__init__(charm, relation_name)
         self.charm = charm
         self.relation_name = relation_name
         self.mount_suffix = mount_suffix
+        self._refresh_events = refresh_events or [self.charm.on.config_changed]
```
gruyaume, line 542: I'm not sure I understand why we observe config changed as a default? Nothing here depends on the requiring charm config right? We may want to use `update status` instead.

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -545,17 +563,20 @@ def _set_unit_egress_subnet(self, relation: ops.Relation, egress_subnet: str):
         """Set the egress_subnet on the relation."""
         relation.data[self.charm.unit]["egress_subnet"] = egress_subnet
 
-    def _on_vault_kv_relation_joined(self, event: ops.RelationJoinedEvent):
+    def _on_vault_kv_relation_joined(self, event: ops.EventBase):
```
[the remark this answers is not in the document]
saltiyazan: This change is initiated by a unit going down then coming up again with a new IP address.
Based on the [documentation](https://juju.is/docs/sdk/config-changed-event) of the `config-changed` event it seems to me that a unit changing its IP will trigger a `config-changed` event.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: This change is initiated by a unit going down then coming up again with a new IP address.
Based on the [documentation](https://juju.is/docs/sdk/config-changed-event) of the `config-changed` event it seems to me that a unit changing its IP will trigger a `config-changed` event.

lib/charms/vault_k8s/v0/vault_kv.py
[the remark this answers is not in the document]
saltiyazan: Agreed. However it is really difficult to come up with a good name here. I am convinced some more in depth refactoring is needed. But I'd keep it out of the scope of this bugfix.

tests/unit/lib/charms/vault_k8s/v0/test_vault_kv.py
```diff
@@ -390,6 +395,15 @@ def test_given_unit_leader_when_unit_joined_then_connected_event_fired_and_all_r
         assert isinstance(event, VaultKvConnectedEvent)
         assert app_relation_data["mount_suffix"] == self.harness.charm.interface.mount_suffix
 
+    @patch("test_vault_kv.VaultKvRequirerCharm._on_connected")
+    def test_given_unit_leader_when_refresh_event_then_connected_event_fired(
+        self, _on_connected
+    ):
+        self.setup_relation()[-1]
```
Kayra1, line 402: We don't need the [-1] here, we're not assigning it to anything.

Kayra1 (approved):
Good job. One minor nit and I would wait for one more approval also.

lib/charms/vault_k8s/v0/vault_kv.py
Kayra1: How do you imagine people would use `refresh_events`? Isn't it enough to just look at `config_changed` since that fires when the IP changes?

saltiyazan: Users don't need to use `refresh_event`, they can specify one depending on their use case. And if they don't we jut use the `config_changed` event.

Kayra1: I'm wondering if they will ever need or want to specify one. What use case is there for it? can you give an example?

tests/unit/lib/charms/vault_k8s/v0/test_vault_kv.py
saltiyazan: Copy paste :D Thanks

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: No, I don't have a use case on top of my mind.
We can silently do `config-changed` in the background if we are convinced that users won't change it.
But I don't mind keeping the flexibility for example if a user doesn't want to trigger this handler on each config change.
what do you think @Kayra1 ?

Kayra1: I think we should remove this option entirely and do it silently with just `config-changed`. It's one less thing for users to worry about and there doesn't seem to be a catch with it that users need to code around or know about.

We can introduce it in the future if someone needs it, but I can't imagine why anyone would want to create a custom event just for one of their libraries either.

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -519,15 +533,19 @@ def __init__(
         charm: ops.CharmBase,
         relation_name: str,
         mount_suffix: str,
+        refresh_events: Optional[List] = None,
```
gruyaume, line 536: The fact that the default is None is somewhat misleading as if it will be set to `config_changed` in this case.  If we want it to be config changed, why not place it here and explain why it's the default in the docstring (which I realise we don't have at the moment)

gruyaume (commented):
I agree with most of this change, it's just the handle_refresh that I think we can improve a bit.

lib/charms/vault_k8s/v0/vault_kv.py
Kayra1: in my opinion this argument can be deleted and only config changed should be observed in the library, but if we really want it to stay I agree that it should default to config changed.

saltiyazan: @gruyaume Do you have any opinions on completely removing it as @Kayra1 suggests? I'd keep it at least for the flexibility it gives but on the other hand I don't have a use case in mind where users would chose a different event.

gruyaume: If we don't have a use case for it, let's not add this unneeded flexibility. And if we keep the config_changed observing, let's mention somewhere why we do it because it's not straightforward why we care about this event.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: Event removed.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: Done.

gruyaume (approved):

PR #365, opened 2024-05-10
feat: Vault autounseal

DanielArndt:
Adds the `vault-autounseal` interface to the `vault-k8s` charm which allows any `vault-k8s` app to integrate with any other app that supports the `vault-autounseal` interface (just `vault-k8s`, for now). This enables autounseal for the requirer side of the relation.

# Caveats / Consequences

The biggest thing to note here, and we can discuss further if necessary, is that there is no way to migrate from autounseal back to manual unseal in this implementation. 

- Manual unseal -> Auto unseal works as expected
- Auto unseal -> Manual unseal works **only if you didn't run the command to migrate** to auto-unseal. i.e. you integrated the two apps, but maybe you made a mistake so you remove the relation without running `vault unseal -migrate ...`

The reason this wasn't straightforward to implement is because to remove auto-unseal, you need to 

1. Add a `disabled = true` key-value to the transit stanza
2. Reboot vault
3. Run the `vault unseal -migrate commands`
4. (optional): remove the transit stanza

In order to implement this we would need to create a new action that stores some state in the charm so that it knows to keep the `disabled = true` key-value in the config. We can't do this after the relation is removed, and we can't do this while the relation is being removed (because we would need to stop and wait for the user to do something).

For now, we keep the auto-unseal key so that it could theoretically be restored, although there is no process to do this yet.

# Usage / Testing

Deploy two vaults:
```console
juju deploy ./vault-k8s_ubuntu-22.04-amd64.charm vault-a --trust -n 1 --resource=vault-image=ghcr.io/canonical/vault:1.15.6
juju deploy ./vault-k8s_ubuntu-22.04-amd64.charm vault-b --trust -n 1 --resource=vault-image=ghcr.io/canonical/vault:1.15.6
```

Initialize and unseal one of the vaults
```
vault_app_ip=...
VAULT_ADDR="https://${vault_app_ip}:8200"
export VAULT_ADDR
vault operator init | tee vault-a-init.txt
export VAULT_TOKEN=$(awk '/Root Token/ { print $4}' vault-a-init.txt)
head vault-a-init.txt -n3 | cut -d' ' -f 4 | xargs -L1 vault operator unseal
```

Authorize the Vault charm
```
juju run vault-a/leader authorize-charm token="$(awk '/Root Token/ { print $4}' vault-a-init.txt)"
```

Integrate the two vaults
```
juju integrate vault-a:vault-autounseal-provides vault-b:vault-autounseal-requires
```

Initialize vault-b
```
vault_app_ip=...
VAULT_ADDR="https://${vault_app_ip}:8200"
export VAULT_ADDR
vault operator init | tee vault-b-init.txt
```

Scale vault-b and watch it autounseal
```
juju scale-application vault-b 2
```

# Checklist:


gruyaume:
> Adds the `vault-autounseal` interface to the `vault-k8s` charm which allows any `vault-k8s` app to integrate with any other app that supports the `vault-autounseal` interface (just `vault-k8s`, for now). This enables autounseal for the requirer side of the relation.
> 
> # Caveats / Consequences
> The biggest thing to note here, and we can discuss further if necessary, is that there is no way to migrate from autounseal back to manual unseal in this implementation.
> 
> * Manual unseal -> Auto unseal works as expected
> * Auto unseal -> Manual unseal works **only if you didn't run the command to migrate** to auto-unseal. i.e. you integrated the two apps, but maybe you made a mistake so you remove the relation without running `vault unseal -migrate ...`
> 
> The reason this wasn't straightforward to implement is because to remove auto-unseal, you need to
> 
> 1. Add a `disabled = true` key-value to the transit stanza
> 2. Reboot vault
> 3. Run the `vault unseal -migrate commands`
> 4. (optional): remove the transit stanza
> 
> In order to implement this we would need to create a new action that stores some state in the charm so that it knows to keep the `disabled = true` key-value in the config. We can't do this after the relation is removed, and we can't do this while the relation is being removed (because we would need to stop and wait for the user to do something).
> 
> For now, we keep the auto-unseal key so that it could theoretically be restored, although there is no process to do this yet.
> 
> # Usage / Testing
> Deploy two vaults:
> 
> ```
> juju deploy ./vault-k8s_ubuntu-22.04-amd64.charm vault-a --trust -n 1 --resource=vault-image=ghcr.io/canonical/vault:1.15.6
> juju deploy ./vault-k8s_ubuntu-22.04-amd64.charm vault-b --trust -n 1 --resource=vault-image=ghcr.io/canonical/vault:1.15.6
> ```
> 
> Initialize and unseal one of the vaults
> 
> ```
> vault_app_ip=...
> VAULT_ADDR="https://${vault_app_ip}:8200"
> export VAULT_ADDR
> vault operator init | tee vault-a-init.txt
> head vault-a-init.txt -n3 | cut -d' ' -f 4 | xargs -L1 vault operator unseal
> ```
> 
> Authorize the Vault charm
> 
> ```
> juju run vault-a/leader authorize-charm token="$(awk '/Root Token/ { print $4}' vault-a-init.txt)"
> ```
> 
> Integrate the two vaults
> 
> ```
> juju integrate vault-a:vault-autounseal-provides vault-b:vault-autounseal-requires
> ```
> 
> Initialize and unseal vault-b
> 
> ```
> vault_app_ip=...
> VAULT_ADDR="https://${vault_app_ip}:8200"
> export VAULT_ADDR
> vault operator init | tee vault-b-init.txt
> head vault-b-init.txt -n3 | cut -d' ' -f 4 | xargs -L1 vault operator unseal
> ```
> 
> Scale vault-b and watch it autounseal
> 
> ```
> juju scale-application vault-b 2
> ```
> 
> # Checklist:
> * [ ]  My code follows the [style guidelines](/CONTRIBUTING.md) of this project
> * [ ]  I have performed a self-review of my own code
> * [ ]  I have made corresponding changes to the documentation
> * [ ]  I have added tests that validate the behaviour of the software
> * [ ]  I validated that new and existing unit tests pass locally with my changes
> * [ ]  Any dependent changes have been merged and published in downstream modules

The guide in the PR description mentions users having to manually unseal Vault B, is this necessary? Reading the tutorial on this, it does not seem necessary

Reference:
- https://developer.hashicorp.com/vault/tutorials/auto-unseal/autounseal-transit

charmcraft.yaml
gruyaume, line 1: I very much like that this behaviour is entirely described through integrations.

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume, line 1: We will need to bump the lib patch version

lib/charms/vault_k8s/v0/vault_client.py
gruyaume, line 1: We will need to bump the lib patch version

lib/charms/vault_k8s/v0/vault_tls.py
gruyaume, line 1: We will need to bump the lib patch version

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -134,6 +136,11 @@ def get_token_data(self) -> Optional[Dict]:
             return None
         return token_data
 
+    @property
+    def token(self):
```
gruyaume, line 140: We are missing return typing

src/templates/vault_transit.hcl.j2
```diff
@@ -0,0 +1,10 @@
+seal "transit" {
+  address         = "{{ vault_addr }}"
+  disable_renewal = "false"
```
gruyaume, line 3: This made me think
- Will the token renew?
- The token expires right?
- If Vault B restarts after token ttl, will it come up?
- Are renewals supported by our implementatin?

src/charm.py
```diff
@@ -162,6 +180,97 @@ def __init__(self, *args):
...
+            self._generate_and_set_autounseal_credentials(relation)
+
+        # FIXME: Don't disable the secrets engine until we have a plan for how
+        # to migrate auto-unseal Vaults back to manual-unseal
+        # if not self.vault_autounseal_provides.get_requirer_requests():
+        #     vault = self._get_active_vault_client()
+        #     if vault is None:
+        #         logger.warning("Vault is not active, cannot disable autounseal transit engine")
+        #         return
+        #     vault.disable_secrets_engine(AUTOUNSEAL_MOUNT_PATH)
```
gruyaume, line 248: Instead of having this as a note in code (or in addition to), can we have a Jira story to make sure this gets discussed during the next backlog grooming

src/charm.py
```diff
@@ -162,6 +180,97 @@ def __init__(self, *args):
...
+
+    def _set_autounseal_relation_data(
+        self, relation: Relation, key_name: str, approle_id: str, approle_secret_id: str
+    ):
```
gruyaume, line 252: Can we add some docstrings? This applies at a couple of places

src/charm.py
```diff
@@ -162,6 +180,97 @@ def __init__(self, *args):
...
+        vault = self._get_active_vault_client()
+        if vault is None:
+            logger.warning("Vault is not active, cannot generate autounseal credentials")
+            return
+
+        vault.enable_secrets_engine(SecretsBackend.TRANSIT, AUTOUNSEAL_MOUNT_PATH)
+
+        # FIXME: We should probably forward the key name in the relation data,
+        # but for now we just infer it on the requirer side, so we ignore it
+        # here.
```
gruyaume, line 225: Humm, why chose the infer approach? I'd suggest providing it. This will allow us to change the scheme in the future if needed

pyproject.toml
```diff
@@ -8,6 +8,7 @@ show_missing = true
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
+filterwarnings = "ignore:::.*.traefik_k8s.*"
```
gruyaume, line 11: Why this specific file? Shouldn't we ignore everything under `lib/`?

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+            policy_name,
+            mount_policy,
+        )
+        return policy_name
+
+    def _get_autounseal_key_name(self, relation_id: int) -> str:
+        """Return the key name for the given relation id."""
+        return str(relation_id)
+
+    def _create_autounseal_key(self, mount_point, relation_id) -> str:
```
gruyaume, line 481: We're missing variable typing

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+
+    def _get_autounseal_policy_name(self, relation_id: int) -> str:
+        """Return the policy name for the given relation id."""
+        return f"charm-autounseal-{relation_id}"
+
+    def _get_autounseal_approle_name(self, relation_id: int) -> str:
+        """Return the approle name for the given relation id."""
+        return f"charm-autounseal-{relation_id}"
+
+    def _configure_autounseal_policy(self, mount: str, relation_id: int, key_name: str):
```
gruyaume, line 459: We're missing typing

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
         raft_config = self._client.sys.read_raft_config()
         return len(raft_config["data"]["config"]["servers"])
+
+    def _get_autounseal_policy_name(self, relation_id: int) -> str:
+        """Return the policy name for the given relation id."""
+        return f"charm-autounseal-{relation_id}"
+
+    def _get_autounseal_approle_name(self, relation_id: int) -> str:
+        """Return the approle name for the given relation id."""
+        return f"charm-autounseal-{relation_id}"
```
gruyaume, line 457: Do those helpers belong in this lib?

gruyaume (changes requested):
This is a preliminary review. Overall I think this looks good, I got some concerns around the token renewal process and the fact that we ask users to unseal their Vault B plus a couple of nits around typing and docstrings.

src/charm.py
DanielArndt: Ooops, I just forgot to remove this comment. We don't infer it anymore.

pyproject.toml
DanielArndt: I added this because I was getting flooded by traefik_k8s warnings from pydantic.

I'm not against widening this, but we would probably want to keep the warnings from our libraries?

I'm not super familiar with the syntax, so I'll need to figure out what we want to do here. I'll probably just remove this from the PR since I don't want it to hold anything up.

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: I guess the alternative would be to make the charm pass in the name of each of them (key name, policy name, approle name)? I felt like this kept the interface a little cleaner, and since it is an internal lib, we're not beholden to it.

Do you think it makes more sense to pass each one of these in as a parameter?

DanielArndt:
> > Initialize and unseal vault-b
> > ```
> > vault_app_ip=...
> > VAULT_ADDR="https://${vault_app_ip}:8200"
> > export VAULT_ADDR
> > vault operator init | tee vault-b-init.txt
> > head vault-b-init.txt -n3 | cut -d' ' -f 4 | xargs -L1 vault operator unseal
> > ```
> 
> Why do we ask users to manually unseal Vault B? Reading the tutorial on this, it does not seem necessary.
> 
> Reference:
> 
> * https://developer.hashicorp.com/vault/tutorials/auto-unseal/autounseal-transit

Sorry, just bad instructions. I was trying to quickly write up a tutorial on what I was doing, but it seems I accidentally included the extra step here because I'm so used to doing that. This doesn't actually need to be done. I'll edit the instructions.

src/charm.py
DanielArndt: To be honest I'm in a bit of a habit of not adding docstrings for private functions unless it's rather complicated.

I feel like some of them are a complete waste (see the `_set_kv_relation_data` or _get_backup_key` docstrings for example). They basically just repeats the function signature and provide nothing of value. It's just more to maintain.

But I get this isn't the place to have the discussion. I will bring it up in the retro.

src/charm.py
DanielArndt: https://warthogs.atlassian.net/browse/TLSENG-242

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/e930a2b82f4c32b824a45b1387644aac5b6724ea

lib/charms/vault_k8s/v0/vault_tls.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/e930a2b82f4c32b824a45b1387644aac5b6724ea

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/e930a2b82f4c32b824a45b1387644aac5b6724ea

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/e930a2b82f4c32b824a45b1387644aac5b6724ea

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/e930a2b82f4c32b824a45b1387644aac5b6724ea

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/e930a2b82f4c32b824a45b1387644aac5b6724ea

src/charm.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/6168f1b1077f8aacf63e8cbe3bcab657d815e71e

src/charm.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/6168f1b1077f8aacf63e8cbe3bcab657d815e71e

pyproject.toml
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/017c5793763c874060d80a05410a0849d504ce2d

src/charm.py
```diff
@@ -948,13 +1080,59 @@ def _generate_vault_config_file(self) -> None:
             node_id=self._node_id,
             retry_joins=retry_joins,
         )
+        content = self._add_transit_stanza_if_needed(content)
```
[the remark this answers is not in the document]
DanielArndt: You're right, it probably does belong there.

src/templates/vault_transit.hcl.j2
```diff
@@ -0,0 +1,10 @@
...
+  key_name        = "{{ key_name }}"
+  mount_path      = "{{ mount_path }}"
+  token           = "{{ vault_token }}"
+  {% if tls_ca_cert %}
```
gruyaume, line 7: Are there cases where we won't have a ca cert for this?

src/charm.py
```diff
@@ -948,13 +1080,59 @@ def _generate_vault_config_file(self) -> None:
...
+        config_content += "\n" + template.render(
+            vault_addr=autounseal_details.address,
+            key_name=autounseal_details.key_name,
+            mount_path=AUTOUNSEAL_MOUNT_PATH,
+            vault_token=vault.token,
+            tls_ca_cert=self.tls.get_tls_file_path_in_workload(File.AUTOUNSEAL_CA),
+        )
+        return config_content
+
+    def _get_transit_requirer_relation(self):
```
gruyaume, line 1133: Return typing is missing

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume, line 1: `VaultAutounsealDestroy` is probably not a great event name, what happened is not that vault autounseal was destroyed, what happened is that the relation to the vault autounseal requirer was broken.

Now what the Vault autounseal provider does when this event occurs is someting else.

src/charm.py
```diff
@@ -162,6 +180,110 @@ def __init__(self, *args):
...
+            self.vault_autounseal_provides.on.vault_autounseal_destroy,
+            self._on_vault_autounseal_destroy,
+        )
+        self.framework.observe(
+            self.vault_autounseal_requires.on.vault_autounseal_provider_removed,
+            self._configure,
+        )
+
+    def _on_vault_autounseal_destroy(self, event: VaultAutounsealDestroy):
+        """Handle the vault-autounseal-destroy event."""
```
gruyaume, line 201: Following the conversation about docstrings, there are likely a couple of them that could be improved and others that could be removed.

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_client.py
Kayra1: I think this belongs in the autounseal library. KV also has its own format for approles, and it doesn't make sense to have both of them here when it's just KV and autounseal that will be using these functions.

lib/charms/vault_k8s/v0/vault_autounseal.py
Kayra1: I like this as a guide to naming these things: https://juju.is/docs/juju/removing-things#heading--removal-terms

Kayra1 (commented):
I'm not fully onboard with the way these changes are structured, and I've made the comments below, but I also think this is a very comprehensive change that is hard to get perfect so I'm ok with not requesting changes and iteratively improving it in the future.

src/charm.py
```diff
@@ -948,13 +1080,59 @@ def _generate_vault_config_file(self) -> None:
...
             existing_content_stringio = self._container.pull(path=VAULT_CONFIG_FILE_PATH)
             existing_content = existing_content_stringio.read()
 
         if not config_file_content_matches(existing_content=existing_content, new_content=content):
             self._push_config_file_to_workload(content=content)
+            if _contains_transit_stanza(existing_content) != _contains_transit_stanza(content):
+                if self._vault_service_is_running():
+                    self._container.restart(self._service_name)
+
+    def _add_transit_stanza_if_needed(self, config_content: str) -> str:
```
saltiyazan, line 1095: I feel like this function is doing a bit more than just adding new config options to the config content it receives, it basically is doing all the configuration that is needed for the autounseal requirer, for example pushing the CA cert and authenticating agains the provider vault.

src/charm.py
```diff
@@ -948,13 +1080,59 @@ def _generate_vault_config_file(self) -> None:
...
         existing_content = ""
         if self._container.exists(path=VAULT_CONFIG_FILE_PATH):
             existing_content_stringio = self._container.pull(path=VAULT_CONFIG_FILE_PATH)
             existing_content = existing_content_stringio.read()
 
         if not config_file_content_matches(existing_content=existing_content, new_content=content):
             self._push_config_file_to_workload(content=content)
+            if _contains_transit_stanza(existing_content) != _contains_transit_stanza(content):
+                if self._vault_service_is_running():
+                    self._container.restart(self._service_name)
```
saltiyazan, line 1093: Why is the restart needed?

saltiyazan: I mean why this does, but a change in the config file not related to the transit stanza doesn't?

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+        policy_name = self._get_autounseal_policy_name(relation_id)
+        self._client.sys.create_or_update_policy(
+            policy_name,
+            mount_policy,
+        )
+        return policy_name
+
+    def _get_autounseal_key_name(self, relation_id: int) -> str:
+        """Return the key name for the given relation id."""
+        return str(relation_id)
```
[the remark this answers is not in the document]
DanielArndt: The idea here is to separate a design decision (the key name) from the rest of the logic. The reason I see this as valuable is because it separates the semantics of the two. The `relation_id` is _not_ the same thing as the key name. We have  decided to name them the same thing for now, but if this changes in the future, we will have one place to modify it, and don't need to go searching for instances of `relation_id` and make sure we replace all the correct instances.

I think this is good practice, but if others don't see the value I will remove.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+
+    def _get_autounseal_policy_name(self, relation_id: int) -> str:
+        """Return the policy name for the given relation id."""
+        return f"charm-autounseal-{relation_id}"
+
+    def _get_autounseal_approle_name(self, relation_id: int) -> str:
+        """Return the approle name for the given relation id."""
+        return f"charm-autounseal-{relation_id}"
+
+    def _configure_autounseal_policy(self, mount: str, relation_id: int, key_name: str) -> str:
```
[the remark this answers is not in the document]
DanielArndt: I'm not against the latter suggestion, but I dislike the first. It doesn't scale. We shouldn't have a function called `_configure_policy` that just grows to contain an ever-expanding list of kwargs that may or may not be relevant to the policy being configured. It doesn't make a nice interface. `mount` is already specific to KV, and if we include the autounseal policy it will also need `key_name`. That doesn't seem right to me. 

It also doesn't allow for things to be inferred without additional input from the caller (such as the policy name, in this case). 

I guess it comes down to how much logic we want to push to the caller. We can remove the kwargs entirely, but then the caller has to fill in the policy parameters themselves. We can not have an autounseal-specific policy function, but then we push this logic to the caller.

I thought the goal of this library was not necessarily just to be an interface to vault, but to avoid repeating logic between the k8s and machine charms. The more we push to the callers, the more work we have to repeat in multiple places.

I'm curious what your thoughts are on this perspective.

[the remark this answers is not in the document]
kayra1: My perspective is that a `policy` is a `policy` no matter what utilizes it. `mount` for example isn't KV specific in general, we can have a custom mount for any policy we want to template. I would agree with you if `_configure_policy` is just a massive switch statement or configuring policies for each thing looked completely different, but in cases where we have 3 different functions and they are just copy and pasted except only 1 line or 1 argument is different, I feel like that's antithetical to what a function is. That is exactly when we should be refactoring multiple functions into one.

However, if we prefer the comfort of that over having a succinct and correct function, let's go with it.

When it comes to sharing code between machine and k8s charms, I don't think this is directly relevant since both charms will use the exact same options whatever we choose. However, in the event where we will need unique functions for either, I think it doesn't make sense to dump specific functions into shared libraries unless those functions are going to be shared verbatim. If we have `func_machine` and `func_k8s` instead of `func(options)`, then there is no point in sharing these functions anyway.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+            policy_name,
+            mount_policy,
+        )
+        return policy_name
+
+    def _get_autounseal_key_name(self, relation_id: int) -> str:
+        """Return the key name for the given relation id."""
+        return str(relation_id)
+
+    def _create_autounseal_key(self, mount_point: str, relation_id: int) -> str:
```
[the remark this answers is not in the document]
DanielArndt: I don't think we can have charm libraries that depend on other charm libraries, so I'm not sure what your suggesting is possible.

As for separation of concerns, I'm not sure I agree but maybe I could be convinced. In this case I don't want a transit key, I want a key for autounseal functionality. The fact that Vault uses transit to facilitate this is an implementation detail, and irrelevant. It could be something else and it shouldn't affect me.

I think we're coming at the vault client from two different perspectives (and I'm not sure I'm right). I've taken it to be a layer between the Vault API and our charms which abstracts the Vault details away so that we don't need to be concerned about them in the charms (and thus can reduce duplication). If it is solely meant to be a translation layer between the Vault API and our charms, I think we'll see a lot more duplication of code between k8s and machine.

Would love to hear some other opinions on this.

src/charm.py
DanielArndt: That's a fair point. It could be refactored to be a ~~little~~ lot cleaner.

src/charm.py
DanielArndt: That's a good question and something I should document.

Changing the seal type requires restarting the node to trigger the changes. It just so happens that transit is the only other seal type we have. This is changing, v1.16 has a beta feature which won't require this.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
kayra1: I think considering the fact that we can't have libraries talking to each other, it's fine to keep it that way.

As far as the Vault client is concerned, my biggest goal with it is to have a vault client that is idempotent and crash avoidant. This means that vault_client's role is a bit different than the rest, but given the constraints around charm libraries this was something we were willing to go with. I do agree that sharing should be another major goal, but once again I don't see how k8s and machine charms will have their code duplicated in any significant way. Preparing the arguments of a function in 2-3 lines is fine to duplicate in my opinion if it means better separation of concerns and easier maintainability (since we see complex logic inline instead of spread out among files).

although that last point is also highly subjective to me.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: To emphasize this, look at [`_configure_autounseal_policy` ](https://github.com/canonical/vault-k8s-operator/blob/017c5793763c874060d80a05410a0849d504ce2d/lib/charms/vault_k8s/v0/vault_client.py#L459-L475)-- We accept both `relation_id` and `key_name` as parameters, even though they are the same (currently). This is because they are semantically different. I could accept just `relation_id`, but then I'd be embedding it in the policy where the `key_name` is, and if the key name ever gets changed the refactor just got a whole lot more complex, searching for all the little cases like this.

Again, I _think_ this is good practice and something I always actually try to make a conscious attempt to do, but I'd love to hear other perspectives.

src/charm.py
```diff
@@ -954,13 +1086,59 @@ def _generate_vault_config_file(self) -> None:
...
+        vault.authenticate(AppRole(autounseal_details.role_id, autounseal_details.secret_id))
+
+        jinja2_environment = Environment(loader=FileSystemLoader(CONFIG_TEMPLATE_DIR_PATH))
+        template = jinja2_environment.get_template(CONFIG_TRANSIT_STANZA_TEMPLATE_NAME)
+
+        config_content += "\n" + template.render(
+            vault_addr=autounseal_details.address,
+            key_name=autounseal_details.key_name,
+            mount_path=AUTOUNSEAL_MOUNT_PATH,
+            vault_token=vault.token,
```
DanielArndt, line 1134: I need to deal with the case where "Vault B" (requirer) cannot contact "Vault A" (provider) for a period of time that is greater than the `token_period`. 

In this case, the token will not be renewed in time, and so it will expire. Unfortunately, as it stands, the charm has no knowledge of the token once it is passed off to the workload, so it can't validate if the token is still valid.

After chatting with @saltiyazan in Madrid, he suggested storing the token in a Juju secret which we can use in the charm to validate and refresh it if necessary. I think this is a good approach and I'll give it a go.

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: @kayra1 as mentioned in [another comment](https://github.com/canonical/vault-k8s-operator/pull/365#discussion_r1625205980), I don't think we can do this without creating cross-library dependencies. Although this might be technically feasible (I'm not sure), I haven't seen it done with any other charm libraries and there's probably a bit of discussion that needs to be had before we start investigating that. This is why, for example, some KV specific stuff already exists in `vault_client` -- because it needs to interact with the Vault cluster directly.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: [If we choose to keep it] I will add a small note to this effect in the docstring.

src/charm.py
DanielArndt: Oops, I didn't notice that this beta feature is only for Enterprise. :cry:

DanielArndt: @saltiyazan Can you let me know if this helps? https://github.com/canonical/vault-k8s-operator/pull/365/commits/87c4ba30a147cabc9bebd7bfafb820386c444c48

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: I spent some time looking at this and I still believe our current implementation of `configure_policy` isn't the best direction forward, but I think I'll do my best to conform for now and suggest an improvement afterwards.

I think that that a refactor is probably too big to include in this PR, and I don't want to fragment the code with two different ways of doing things.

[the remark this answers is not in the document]
kayra1: I had done the PR to move to that [here](https://github.com/canonical/vault-k8s-operator/pull/234/files), which received no comments there. If we as a team decide otherwise, I have no problem switching back.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+        key_name = self._get_autounseal_key_name(relation_id)
+        response = self._client.secrets.transit.create_key(mount_point=mount_point, name=key_name)
+        logging.debug(f"Created a new autounseal key: {response}")
+        return key_name
+
+    def _destroy_autounseal_key(self, mount_point, key_name):
+        """Destroy the autounseal key."""
+        self._client.secrets.transit.delete_key(mount_point=mount_point, name=key_name)
+
+    def destroy_autounseal_credentials(self, relation_id: int, mount: str) -> None:
```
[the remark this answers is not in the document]
kayra1: responded there but in short, lets keep it this way if we want to avoid cross lib dependencies.

Just as an aside, I didn't think cross lib dependencies would be an issue because at some point [Loki was doing it](https://github.com/canonical/loki-k8s-operator/commit/f8601e7a058ccf52c47cd86178aa8649b7f02893#diff-ad7051b5ca68befcdb77d638fcfa686fa902ae8089c8e72cea3c222df01a4597).

But it looks like they're moving away from that and publish a `cosl` package in pypi to share library dependencies now.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: It's not about "switching back", it's about an alternative way to move forward. There's nothing wrong with the way it was implemented at the time, it just makes some assumptions that make it hard to re-use in new use cases.

I tried to make the code conform last night and although `configure_policy()` was a little unfortunate for the reasons mentioned above, `configure_approle()` gets messy too because we've made some assumptions about the policy as well. For example, we don't want to set a `token_ttl`, or `token_max_ttl`. We want to set `token_period` instead. Can we start adding some branching to `configure_approle()` based on our business logic? Absolutely. But that gets really difficult to follow or keep correct as this continues to absorb new use cases.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: > my biggest goal with it is to have a vault client that is idempotent and crash avoidant.

I think this is a good point and something we should continue to strive for. I'm going to make a follow-up ticket for this discussion, and I think one way of emphasising this through architecture would be to have two separate classes in the library. One that provides the interface to Vault with all our handling and parsing on top of it (the fault resistant, idempotent interface), and the other that wraps it with our business logic. This is based on some discussion I had with @ghislainbourgeois yesterday, so I don't want to make it sound like it was my idea.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: yeah I had some discussion yesterday with @ghislainbourgeois and he mentioned it's possible, but the experience isn't great. Probably something we still want to discuss before we start.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: Deferring this discussion to a spec: https://warthogs.atlassian.net/browse/TLSENG-269

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: Deferring this discussion to a spec: https://warthogs.atlassian.net/browse/TLSENG-269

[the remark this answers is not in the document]
kayra1: > I tried to make the code conform last night and although `configure_policy()` was a little unfortunate for the reasons mentioned above, `configure_approle()` gets messy too because we've made some assumptions about the policy as well. For example, we don't want to set a `token_ttl`, or `token_max_ttl`. We want to set `token_period` instead. Can we start adding some branching to `configure_approle()` based on our business logic? Absolutely. But that gets really difficult to follow or keep correct as this continues to absorb new use cases.

It's more about `token_ttl: Optional = None`. Branching statements that increase complexity and separating functions by copy pasting most of it and changing a couple of lines is a false dichotomy. There are many ways to make functions more dynamic without branching if statements, and many ways to split the same functionality into two different most common use-case functions with minimal arguments. 

In this case I believe it's preferable to share the function. If we need to add an extra argument so you can still have the function signature you want, but we need to add 2 `"60s"`'s into other invocations, that's great. In fact adding 60s's to other invocations is better than having 60s as default, it's more explicit so in my eyes that's a win-win situation. If we make two functions that have different arguments to the same hvac call, that's just 1 win. And in this case, I believe that two functions share more similarities than not.

Feel free to resolve this when you read it, I agree that a spec is probably a better place to talk about this, and there's really no need to hold this PR up.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
kayra1: I'm not opposed to that idea at all. also, talked a little about this in my other comment [above](https://github.com/canonical/vault-k8s-operator/pull/365#discussion_r1627961251) just now. feel free to resolve this when you read it.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: Makes sense. Unfortunately it needs to be here as this is where the vault interaction is that relies on it (unless we create a cross library dependency the other way, too).

Charm libraries are definitely pushing us to some less-than-ideal approaches. That being said, perhaps the outcome of https://warthogs.atlassian.net/browse/TLSENG-269 will improve this a bit.

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: I'm not suggesting copy/pasting or repeating parts anywhere. I could have factored out the common parts for sure, and that's direction I would like to move. But, as they're written, they're not easy to re-use. If we continue to add more variables every time we want different values, this library effectively becomes a thin wrapper around hvac, with the exact same parameters, just surrounded by a try..except block.

Keep in mind, also, that a majority of this was written before your refactor as I started to work on this back in February before I had to put it aside for Manual Unseal. My bad for not updating it in consideration of the new changes, but hopefully that provides some context.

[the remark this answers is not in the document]
kayra1: > this library effectively becomes a thin wrapper around hvac, with the exact same parameters, just surrounded by a try..except block.

you say that like it's a bad thing :D I think a bunch of my comments really stems from the fact that I do want this. and I only commented instead of asking for changes because I'm aware that this PR is not the right place to tackle this. In the future I think I'm going to avoid these types of comments and open a ticket for us to talk about it in the backlog grooming meeting.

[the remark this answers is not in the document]
DanielArndt: It was great conversation, so I appreciate it nonetheless.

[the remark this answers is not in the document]
kayra1: always love talking with you

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: I'm going to leave this as-is for now. I don't want to push the charm to having too much knowledge about the internal workings of Vault if it doesn't need to be there until we've had some discussion about https://warthogs.atlassian.net/browse/TLSENG-269

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -416,3 +447,100 @@ def get_num_raft_peers(self) -> int:
...
+        """Create/update a policy within vault to use the transit key."""
+        mount_policy = textwrap.dedent(f"""
+        path "{mount}/encrypt/{key_name}" {{
+            capabilities = ["update"]
+        }}
+
+        path "{mount}/decrypt/{key_name}" {{
+            capabilities = ["update"]
+        }}
+        """)
```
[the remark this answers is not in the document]
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/e9f318763e6f85ed946049a862f15f5bb94885f2

lib/charms/vault_k8s/v0/vault_client.py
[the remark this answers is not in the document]
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/e9f318763e6f85ed946049a862f15f5bb94885f2

src/templates/vault_transit.hcl.j2
DanielArndt: No, not anymore. This can be removed.

src/charm.py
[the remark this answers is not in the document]
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/0a90837bbcdef6e42799e4e45129a758b899699f

src/templates/vault_transit.hcl.j2
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/0a90837bbcdef6e42799e4e45129a758b899699f

src/charm.py
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/0a90837bbcdef6e42799e4e45129a758b899699f

src/charm.py
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/0a90837bbcdef6e42799e4e45129a758b899699f

lib/charms/vault_k8s/v0/vault_autounseal.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/26d8322af707e9a05ab70a4b408d4f70a0f6b62d

src/charm.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/ea12fcf5c8462cb0adc3603db5a8b1c71d6086ca

saltiyazan:
The CI is failing because of a new test case, a fix is attempted here https://github.com/canonical/vault-k8s-operator/pull/388

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -422,52 +420,84 @@ def __init__(self, charm: ops.CharmBase, relation_name: str):
...
+                details.secret_id,
+                details.ca_certificate,
             )
 
     def _on_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
-        self.on.vault_autounseal_provider_removed.emit()
+        self.on.vault_autounseal_provider_relation_broken.emit()
 
-    def get_details(self) -> AutounsealDetails:
+    def get_details(self) -> AutounsealDetails | None:
```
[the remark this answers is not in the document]
DanielArndt: I lean toward `| None` for three reasons:

1. Optional was a poor choice of words. It doesn't fit well in all contexts and it conflicts with the already established meaning of an optional argument (now you can have an optional argument that isn't optional :laughing: )
2. It's easier to read (subjective)
3. It's the new way, and new things are always better (totally objective :wink: ). OK honestly this is a repeat of #1, but they moved this direction for a reason.

Anyway, if you're suggesting either changing them all, or changing these ones, I'll just change this to `Optional` :laughing:

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -208,18 +211,38 @@ def restore(self, snapshot: Dict[str, Any]) -> None:
...
 class VaultAutounsealRequireEvents(ops.ObjectEvents):
     """Events raised by the vault-autounseal relation on the requirer side."""
 
     vault_autounseal_details_ready = ops.EventSource(VaultAutounsealDetailsReadyEvent)
-    vault_autounseal_provider_removed = ops.EventSource(VaultAutounsealProviderRemoved)
+    vault_autounseal_provider_relation_broken = ops.EventSource(VaultAutounsealProviderRemoved)
+
+
+@dataclass
+class AutounsealDetails:
```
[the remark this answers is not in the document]
DanielArndt: I'm not sure calling the class `Autounseal` and `Approle` makes sense. These classes don't have or support any functionality for either of those things, and they're merely a container for the information.

Perhaps they could be renamed `AutounsealConfigurationProperties` and `ApproleLoginDetails` or something, but I think removing the `Details` part makes it more confusing, personally. I was under the impression this is a fairly common pattern for these types of classes, but maybe I'm incorrect.

src/charm.py
```diff
@@ -946,6 +1110,71 @@ def _cluster_address(self) -> str:
...
+        self.tls.push_autounseal_ca_cert(autounseal_details.ca_certificate)
+
+        vault = Vault(
+            url=autounseal_details.address,
+            ca_cert_path=self.tls.get_tls_file_path_in_charm(File.AUTOUNSEAL_CA),
+        )
+        existing_token = self._get_juju_secret_field(AUTOUNSEAL_TOKEN_SECRET_LABEL, "token")
+        if not existing_token or not vault.authenticate(Token(existing_token)):
+            vault.authenticate(AppRole(autounseal_details.role_id, autounseal_details.secret_id))
+            self._set_juju_secret(AUTOUNSEAL_TOKEN_SECRET_LABEL, {"token": vault.token})
```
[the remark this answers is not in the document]
DanielArndt: Could definitely move this to a new method.

This part of the code generates a new token if either

1. There is no existing token, or
2. The current token is no longer valid. This is the case where it failed to call home early enough, and so the token expired.

lib/charms/vault_k8s/v0/vault_autounseal.py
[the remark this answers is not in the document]
gruyaume: I really don't mind which we chose, I only mind that there are 2 different ways to do the same thing in 1 repo.

src/charm.py
```diff
@@ -160,6 +191,116 @@ def __init__(self, *args):
...
+        if vault is None:
+            logger.warning("Vault is not active, cannot disable vault autounseal")
+            return
+        vault.destroy_autounseal_credentials(event.relation.id, AUTOUNSEAL_MOUNT_PATH)
+
+    def _on_vault_autounseal_requirer_relation_created(
+        self, event: VaultAutounsealRequirerRelationCreated
+    ):
+        """Generate and set the auto-unseal credentials for the requirer."""
+        self._generate_and_set_autounseal_credentials(event.relation)
```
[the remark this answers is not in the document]
DanielArndt: We can, and I guess that's more in-line with the holistic pattern so I'll do that.

lib/charms/vault_k8s/v0/vault_autounseal.py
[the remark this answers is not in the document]
gruyaume: This is true of all dataclasses though isn't it? Should add the `Details` suffix to every dataclass? It may be a common pattern, I really don't know.

src/charm.py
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/a18ace011a04222a4fa35c6c24686b9051112506

lib/charms/vault_k8s/v0/vault_autounseal.py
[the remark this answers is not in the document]
DanielArndt: > This is true of all dataclasses though isn't it? Should add the `Details` suffix to every dataclass?

That's a good point, let me revise my argument a little  -- I think it's important to consider what these classes represent. They don't represent Autounseal or Approle. Just the details surrounding their usage.

In most examples of dataclasses I find things like `Person`, `Rectangle`, `InventoryItem`, etc. Those all make sense because they actually represent those things in the context. But the containers we're discussing do not. They're only used to pass chunks of data.  If I saw an object of type `Autounseal`, I would expect it to represent some manifestation of auto-unseal, not merely the configuration details needed to set it up in the vault config.

lib/charms/vault_k8s/v0/vault_autounseal.py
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/365/commits/5c95fc7280cc0980d6a4cf6a05c0d6435b01da14

src/charm.py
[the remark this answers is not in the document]
DanielArndt: > Plus are there considerations related to the token ttl?

Missed this part of the question. No, the default configuration for the token are set on the AppRole.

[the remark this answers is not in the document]
DanielArndt: Addressed in https://github.com/canonical/vault-k8s-operator/pull/365/commits/c6c7591e0049250f093fceeaefbfc34b7522039c

lib/charms/vault_k8s/v0/vault_autounseal.py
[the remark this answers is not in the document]
DanielArndt: @gruyaume I don't have a super strong opinion here, but enough to push back a little. If I still haven't convinced you I'll change the names. Let me know.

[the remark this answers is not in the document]
gruyaume: Good with me!

gruyaume (approved):

PR #373, opened 2024-05-27
fix: Uses release branch tag in the CI

saltiyazan:
# Description

Preparing release branch `release-1.15`.

# Checklist:


kayra1:
im not too familiar with this, does this PR create a new workflow only for this branch?

saltiyazan:
> im not too familiar with this, does this PR create a new workflow only for this branch?

Yes, it will trigger the CI when we create a PR on this branch. And this PR already shows that, the CI is running although we are not pushing anything to main.

.github/workflows/codeql-analysis.yml
```diff
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "release-1.15" ]
```
ghislainbourgeois, line 16: Maybe instead of having different files per branch listing only the branch we are on, we could use a pattern like `release-*` and keep `main` in the list? Then we could take decision like what channel to push to based on the name of the branch.

ghislainbourgeois (changes requested):

.github/workflows/codeql-analysis.yml
saltiyazan: Brilliant, thanks.
Done in #424 please take a look

saltiyazan: The same change added here too for testing purposes, but when #424 is merged there will be a rebase here

saltiyazan:
I added dependabot config specific for this branch. Let me know in case of any objections @ghislainbourgeois

.github/workflows/codeql-analysis.yml
saltiyazan: Rebased

ghislainbourgeois (approved):

PR #382, opened 2024-06-03
fix: freeze dependencies in vault kv requirer

gruyaume:
# Description

Addresses issues in `main` that made integration tests fail:
1. Freeze dependencies in kv requirer charm: There was a silent dependency update in vault kv requirer that caused ops and ops-scenario to be out of sync. Using pip-compile we can use compatible versions of the two.
2. Revert kubernetes_charm_libraries lib patch change: The latest version of the lib creates a bad k8s service name (`self.service_name = service_name or self._app`) and we revert to the last working version of the lib. 

# Checklist:


saltiyazan (approved):
Awesome! Thanks for fixing this.

Are you going to raise an issue against the bug in the kubernetes_service_patch lib?

I still have some issues locally, but it's probably just my env.

gruyaume:
> Awesome! Thanks for fixing this.
> 
> Are you going to raise an issue against the bug in the kubernetes_service_patch lib?
> 
> I still have some issues locally, but it's probably just my env.

We can get rid of the lib altogether in Vault K8s by using the recent `self.unit.set_ports` feature in ops

PR #386, opened 2024-06-06
fix: Sets the latest issuer to be the default issuer in PKI

saltiyazan:
# Description

Since v1.11 Vault now creates a new issuer whenever a CA is added in PKI. However the default issuer is not updated by default, so when the charm checks the current intermediate CA it will get the one of the default (first) issuer, causing the charm to generate a new CSR and a new CA and push that to Vault over and over again once the common name is changed.

Although the charm would work as expected and provide certificates under the new common name, if the charm is deployed for long enough one would see a large number of CAs created.

This PR fixes that issue, now when a new common name is set and new CA is created we make sure that the new issuer is the default. This is achieved by setting the default issuer to the latest issuer after the creation of the first issuer.

**Note for reviewer**:
This is only a temporary fix to the phenomena described above, but the way to handle the change of CAs and Issuers should be agreed upon as part of the effort to define how to handle CA expiry and change in the TLS provider charms.

# Checklist:

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -48,6 +48,9 @@ TLS_CERTIFICATES_LIB_PATH = "charms.tls_certificates_interface.v3.tls_certificat
 VAULT_KV_RELATION_NAME = "vault-kv"
 TLS_CERTIFICATES_PKI_RELATION_NAME = "tls-certificates-pki"
 VAULT_KV_REQUIRER_APPLICATION_NAME = "vault-kv-requirer"
+PKI_MOUNT = "charm-pki"
+PKI_ROLE_NAME = "charm"
+APPROLE_ROLE_NAME = "charm"
 
 
 def read_file(path: str) -> str:
@@ -536,10 +539,10 @@ class TestCharm(unittest.TestCase):
             policy_name=CHARM_POLICY_NAME, policy_path=CHARM_POLICY_PATH
         )
         self.mock_vault.configure_approle.assert_called_once_with(
-            role_name="charm", policies=[CHARM_POLICY_NAME, "default"], cidrs=["10.0.0.10/24"]
+            role_name=APPROLE_ROLE_NAME, policies=[CHARM_POLICY_NAME, "default"], cidrs=["10.0.0.10/24"]
         )
         self.mock_vault.generate_role_secret_id.assert_called_once_with(
-            name="charm", cidrs=["10.0.0.10/24"]
+            name=APPROLE_ROLE_NAME, cidrs=["10.0.0.10/24"]
         )
 
         secret_content = self.harness.model.get_secret(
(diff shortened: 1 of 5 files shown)


tests/unit/test_charm.py
```diff
@@ -1753,3 +1756,66 @@ def test_given_vault_available_when_vault_pki_certificate_creation_request_then_
...
+        self.harness.charm._on_tls_certificate_pki_certificate_available(event=event)
+
+        self.mock_vault.set_pki_intermediate_ca_certificate.assert_called_with(
+            certificate=certificate,
+            mount=PKI_MOUNT,
+        )
+        self.mock_vault.create_or_update_pki_charm_role.assert_called_with(
+            allowed_domains="vault", mount=PKI_MOUNT, role=PKI_ROLE_NAME
+        )
+        self.mock_vault.make_latest_pki_issuer_default.assert_called_with(mount=PKI_MOUNT)
```
ghislainbourgeois, line 1821: Is there a more concrete way we could test this? This seems a bit to indirect and not really validating that it has the right issuer, and as this is a temporary fix, the next one will probably require changing this test again.

ghislainbourgeois (changes requested):

tests/unit/test_charm.py
saltiyazan: Whether this is a temporary change or not depends on design decisions on how to handle CA changes not on implementation decisions, is it a problem in that case that the tests might change?

As for testing this in a more concrete way I added tests for the `make_latest_pki_issuer_default` in the client tests.
I don't know if that is enough but here in the charm tests we don't seem to validate that changes made by the client really took effect. For example we don't test that the role desired by `create_or_update_pki_charm_role` has been actually created. For us to know which issuer is being used we need to read the config on `/pki/config/issuers`

Anyway I do see your point and I agree with it, but I don't think it is unique to this function.

kayra1: How about an integration test step [here](https://github.com/canonical/vault-k8s-operator/blob/4e412beb4ceb1aa73aebd7e10b30464b47fc7fa0/tests/integration/test_integration.py#L386)? just add a change config in between and validate with a new requirer?

kayra1 (commented):
Not too familiar with this, but [this](https://developer.hashicorp.com/vault/tutorials/secrets-management/pki-engine#step-11-sunset-defunct-root-ca) tutorial from vault also explains how to sunset old CA's. maybe we're interested in doing that too?

saltiyazan:
> Not too familiar with this, but [this](https://developer.hashicorp.com/vault/tutorials/secrets-management/pki-engine#step-11-sunset-defunct-root-ca) tutorial from vault also explains how to sunset old CA's. maybe we're interested in doing that too?

Yes, that is something we want to care about. But since now we don't really have a strategy I would not include it in the scope of this PR I'd prefer to figure that out first.
Now a new common name will create a new CA and update the role and update the allowed_domains in that role, so effectively the old CA won't be able to issue certificate because of the change in the role.

tests/unit/test_charm.py
saltiyazan: Good point @kayra1 
I added integration tests.

PR #388, opened 2024-06-07
fix: Validate if issuers config was already updated before updating it

saltiyazan:
# Description

- We validate that the issuers config is not yet updated before updating it again. This should prevent us from writing to that config more than once avoiding any unexpected behaviour.
- The integration test changes introduced in #386 made the CI unstable. The changes in the issuers config might not be immediately visible in vault, for that in this PR we wait until all units are active idle (that indicates that the Vault API is available again).

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_client.py
+++ b/lib/charms/vault_k8s/v0/vault_client.py
@@ -469,21 +469,19 @@ class Vault:
         except (InvalidPath, KeyError) as e:
             logger.error("No issuers found on the specified path: %s", e)
             raise VaultClientError("No issuers found on the specified path.")
+        issuers_config = self._client.read(path=f"{mount}/config/issuers")
         try:
-            issuers_config = self._client.read(path=f"{mount}/config/issuers")
-            if issuers_config["data"]["default_follows_latest_issuer"] == True:
-                logger.debug("Config is already set")
-                return
-        except:
-            pass
-        logger.debug("Updating issuers config")
-        self._client.write_data(
-            path=f"{mount}/config/issuers",
-            data={
-                "default_follows_latest_issuer": True,
-                "default": first_issuer,
-            },
-        )
+            if issuers_config and not issuers_config["data"]["default_follows_latest_issuer"]:  # type: ignore -- bad type hint in stubs
+                logger.debug("Updating issuers config")
+                self._client.write_data(
+                    path=f"{mount}/config/issuers",
+                    data={
+                        "default_follows_latest_issuer": True,
+                        "default": first_issuer,
+                    },
+                )
+        except (TypeError, KeyError):
+            logger.error("Issuers config is not yet created")
 
     def _get_autounseal_policy_name(self, relation_id: int) -> str:
         """Return the policy name for the given relation id."""
--- a/lib/charms/vault_k8s/v0/vault_client.py
+++ b/lib/charms/vault_k8s/v0/vault_client.py
@@ -469,6 +469,14 @@ class Vault:
         except (InvalidPath, KeyError) as e:
             logger.error("No issuers found on the specified path: %s", e)
             raise VaultClientError("No issuers found on the specified path.")
+        try:
+            issuers_config = self._client.read(path=f"{mount}/config/issuers")
+            if issuers_config["data"]["default_follows_latest_issuer"] == True:
+                logger.debug("Config is already set")
+                return
+        except:
+            pass
+        logger.debug("Updating issuers config")
         self._client.write_data(
             path=f"{mount}/config/issuers",
             data={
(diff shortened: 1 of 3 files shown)


DanielArndt:
This might be because `wait_for_idle` is extremely easy to misuse, especially when dealing with multiple applications.

By default, `wait_for_at_least_units` is 1, which means (as I understand) that as soon as _any_ one unit that belongs to the applications listed is "active", it will move forward.

So, we may want to bump the number of units to wait for to (num vault units + 1 for pki requirer) to make sure all units are ready. If we really only need one vault unit to be ready, I think we'd need to split them into two separate waits.

saltiyazan:
> This might be because `wait_for_idle` is extremely easy to misuse, especially when dealing with multiple applications.
> 
> By default, `wait_for_at_least_units` is 1, which means (as I understand) that as soon as _any_ one unit that belongs to the applications listed is "active", it will move forward.
> 
> So, we may want to bump the number of units to wait for to (num vault units + 1 for pki requirer) to make sure all units are ready. If we really only need one vault unit to be ready, I think we'd need to split them into two separate waits.

After some testing, yes that should work, however I don't see right now the connection between the status of the unit and what config is applied in Vault. I propose having both to avoid any unstability in the CI.

DanielArndt:
> I don't see right now the connection between the status of the unit and what config is applied in Vault

If we don't wait for it to settle after calling `vault_app.set_config()`, we may be hitting the API before it has time to consider the changed. Calling `wait_for_idle()` with the PKI requirer app in the parameters means that we don't wait at all because we haven't done anything to require work on the PKI requirer app, so it is already idle and we immediately return.

There may be some cross test issue too, since in the test above we make the same wait call, and we're not actually waiting for both charms to finish dealing with the integration.

> I propose having both to avoid any unstability in the CI.

Personally, I'd rather lean the other direction. Lets not add complication unless we're sure we need it. If we don't definitively know the cause and it seems to work after correcting this `wait_for_idle()` stuff, I'd let it simmer for a bit and see if it re-emerges.

saltiyazan:
> > I don't see right now the connection between the status of the unit and what config is applied in Vault
> 
> If we don't wait for it to settle after calling `vault_app.set_config()`, we may be hitting the API before it has time to consider the changes. Calling `wait_for_idle()` with the PKI requirer app in the parameters means that we don't wait at all because we haven't done anything to require work on the PKI requirer app, so it is already idle and we immediately return.
> 
> There may be some cross test issue too, since in the test above we make the same wait call, and we're not actually waiting for both charms to finish dealing with the integration.
> 
> > I propose having both to avoid any unstability in the CI.
> 
> Personally, I'd rather lean the other direction. Lets not add complication unless we're sure we need it. If we don't definitively know the cause and it seems to work after correcting this `wait_for_idle()` stuff, I'd let it simmer for a bit and see if it re-emerges.

Okay, I see your point, I reverted those changes and kept the ones for waiting for the units to be active idle. We can revisit if the problem occurs.
Can you please take a look? @DanielArndt

saltiyazan:
> > > I don't see right now the connection between the status of the unit and what config is applied in Vault
> > 
> > 
> > If we don't wait for it to settle after calling `vault_app.set_config()`, we may be hitting the API before it has time to consider the changes. Calling `wait_for_idle()` with the PKI requirer app in the parameters means that we don't wait at all because we haven't done anything to require work on the PKI requirer app, so it is already idle and we immediately return.
> > There may be some cross test issue too, since in the test above we make the same wait call, and we're not actually waiting for both charms to finish dealing with the integration.
> > > I propose having both to avoid any unstability in the CI.
> > 
> > 
> > Personally, I'd rather lean the other direction. Lets not add complication unless we're sure we need it. If we don't definitively know the cause and it seems to work after correcting this `wait_for_idle()` stuff, I'd let it simmer for a bit and see if it re-emerges.
> 
> Okay, I see your point, I reverted those changes and kept the ones for waiting for the units to be active idle. We can revisit if the problem occurs. Can you please take a look? @DanielArndt

The CI just failed, I'm not sure if I'm doing something wrong based on your description or if this won't be enough.
Check https://github.com/canonical/vault-k8s-operator/actions/runs/9463126946/job/26068340370?pr=388

DanielArndt:
Hmm... still broken even with the wait included again... so weird. This might be worthy of some additional investigation. It seems like we're not accounting for something.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -469,13 +469,19 @@ def make_latest_pki_issuer_default(self, mount: str) -> None:
...
-        self._client.write_data(
-            path=f"{mount}/config/issuers",
-            data={
-                "default_follows_latest_issuer": True,
-                "default": first_issuer,
-            },
-        )
+        try:
+            issuers_config = self._client.read(path=f"{mount}/config/issuers")
+            if issuers_config and not issuers_config["data"]["default_follows_latest_issuer"]:  # type: ignore -- bad type hint in stubs
```
DanielArndt, line 474: I'm assuming by the way this is written, `issuers_config` exists as some default value even before we write anything?

saltiyazan: Yes, my intention is to wait until Vault creates the config then we modify it.

DanielArndt (approved):

DanielArndt:
I'm going to merge this because it will help me publish the latest library changes and get them into the machine charm.

PR #428, opened 2024-07-16
feat: enable dead server pruning

kayra1:
# Description

This PR introduces a change that [configures the raft backend](https://developer.hashicorp.com/vault/api-docs/system/storage/raftautopilot#set-configuration) when the charm is being authorized. These configuration options do not have a relevant field in the vault.hcl config file, so they have to be executed (relatively) manually.

Vault only allows a minimum of 3 for the minimum quorum for automatic dead server pruning. This means that dead server 2 and 3 will never be cleaned. On top of that, we already do our best to update the vault config file and remove the peers that are not planned to be there, so this is really a plan C thing.

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_client.py
+++ b/lib/charms/vault_k8s/v0/vault_client.py
@@ -26,7 +26,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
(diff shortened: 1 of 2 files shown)


DanielArndt:
Blocked by #429

kayra1:
> Blocked by #429

Looks like its working now

gruyaume (approved):

Issue #415, opened 2024-07-03
ModelError: ERROR cannot ensure service account [...] Internal error occurred: resource quota evaluation timed out


DanielArndt:
Reported by @alesstimec

### Bug Description

This error [seems to come from k8s](https://github.com/kubernetes/kubernetes/issues/67531), and it looks like a resource (etcd? ceph?) is just taking too long to respond. This aligns with the ceph outage mentioned at the time.

### To Reproduce

Unknown

### Environment

https://github.com/canonical/vault-k8s-operator/issues/407#issuecomment-2194556611

>  this was in a live model on our prodstack. I believe there might have been an ongoing ceph outage at the time, so the entire controller was a bit sluggish to respond. To resolve the issue i had to redeploy vault once controller was back on track.

### Relevant log output

```shell
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-vault-0/charm/./src/charm.py", line 1575, in <module>
    main(VaultCharm)
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/main.py", line 548, in main
    manager.run()
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/main.py", line 527, in run
    self._emit()
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/main.py", line 516, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/main.py", line 147, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/framework.py", line 348, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/framework.py", line 860, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-vault-0/charm/venv/ops/framework.py", line 950, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-vault-0/charm/./src/charm.py", line 408, in _configure
    self._configure_pki_secrets_engine()
  File "/var/lib/juju/agents/unit-vault-0/charm/./src/charm.py", line 485, in _configure_pki_secrets_engine
    vault = self._get_active_vault_client()
  File "/var/lib/juju/agents/unit-vault-0/charm/./src/charm.py", line 1377, in _get_active_vault_client
[...]

PR #434, opened 2024-07-19
refactor: Remove secret usage from Vault PKI implementation for CA CSR

DanielArndt:
Part of the solution for: #415 
 
Currently, we store the CSR in a secret so that we can compare the certificate that is returned with the one we sent. However, we already store this CSR in the relation databag, and we are only ever requesting a single certificate.

Removing the dependence on secrets will simplify the solution for #415 because we won’t need to consider as many cases in the event that adding or updating the secret fails. Furthermore, it simplifies our code.

# Checklist:

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -85,7 +85,6 @@ KV_RELATION_NAME = "vault-kv"
 KV_SECRET_PREFIX = "kv-creds-"
 LOG_FORWARDING_RELATION_NAME = "logging"
 PEER_RELATION_NAME = "vault-peers"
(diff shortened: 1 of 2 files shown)


gruyaume (changes requested):
Can you improve the PR description, only looking at the code changes does not make it obvious why we are doing them.

DanielArndt:
@gruyaume yeah, sorry, I added it to the Jira ticket but forgot to move it here.

gruyaume (approved):

PR #448, opened 2024-07-29
fix: Adds a workaround to avoid changing relation data when kv secret id formatting changes

saltiyazan:
# Description

fixes https://github.com/canonical/vault-operator/issues/243

In case of cross model relations when the kv secret is created it has the format `secret://modeluuid/secretid`.
However when the secret is updated using the `set_content` function the id format becomes `secret://secretid` which is not usable in cross model relation, we add a workaround not to update the relation data with this new format.
A bug will be reported in Juju.

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_kv.py
+++ b/lib/charms/vault_k8s/v0/vault_kv.py
@@ -379,7 +379,9 @@ class VaultKvProvides(ops.Object):
                 nonce,
             )
             return
-        standard_secret_id = secret.id[len("secret:"):] if secret.id.startswith("secret:") else secret.id
+        standard_secret_id = (
+            secret.id[len("secret:") :] if secret.id.startswith("secret:") else secret.id
+        )
         if standard_secret_id in credentials.get(nonce, ""):
             logger.debug(
                 "Secret id has not changed, not updating the relation '%s:%d' for nonce %r",
--- a/lib/charms/vault_k8s/v0/vault_kv.py
+++ b/lib/charms/vault_k8s/v0/vault_kv.py
@@ -133,7 +133,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 10
 
 PYDEPS = ["pydantic", "pytest-interface-tester"]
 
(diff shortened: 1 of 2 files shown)


lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -379,6 +379,17 @@ def set_unit_credentials(
                 nonce,
             )
             return
+        standard_secret_id = (
+            secret.id[len("secret:") :] if secret.id.startswith("secret:") else secret.id
+        )
```
gruyaume, line 384: Please feel free to add comments in the docstring or right here related to why we are doing this thing. I know it's explained in the PR description, but in 6 months if someone reads this code it may not be obvious.

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -379,6 +379,17 @@ def set_unit_credentials(
...
+        if standard_secret_id in credentials.get(nonce, ""):
+            logger.debug(
+                "Secret id has not changed, not updating the relation '%s:%d' for nonce %r",
+                relation.name,
+                relation.id,
+                nonce,
+            )
+            return
         credentials[nonce] = secret.id
```
gruyaume, line 393: Shouldn't it be `credentials[nonce] = standard_secret_id`?

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: No, it is correct the way it is.
We don't want to store the standard format.
The following is happening:
- The secret is created with an ID of the format `secret://model_uuid/standard_secret_id`
- ID is stored using `credentials[nonce] = secret.id`
- When the secret is updated later juju might return and ID with the format `secret://standard_secret_id` or `standard_secret_id` which can't be used to access the secret from a different model.

So with this workaround we keep the `secret://model_uuid/standard_secret_id` format.

saltiyazan:
Closing in favour of a better solution. https://github.com/canonical/vault-operator/pull/255/files
Getting the secret by ID instead of label should solve this problem, and that solution does not look very much as a workaround. (We have a similar mechanism in place for the k8s charm).
Reporting the bug in juju is still required however.

saltiyazan:
Closing in favour of a better solution. https://github.com/canonical/vault-operator/pull/255/files
Getting the secret by ID instead of label should solve this problem, and that solution does not look very much as a workaround. (We have a similar mechanism in place for the k8s charm).
Reporting the bug in juju is still required however.

PR #450, opened 2024-08-01
fix: Uses egress subnet list to configure approle

saltiyazan:
# Description

**Note**: This is a breaking change on the provider side, but since this is handled in the same PR in Vault I think it should be okay. No breaking changes on the requirer side by allowing it to set the egress subnet as string or as list.

Instead of using the `interfaces` of the network binding to configure the approle in vault we use the list of `egress_subnets`.
Juju describes the interfaces as:
```
A list of network interface details. This includes the information
    about how the application should be configured (for example, what IP
    addresses should be bound to).
```

While it describes the `egress_subnets` to be:
```
A list of networks representing the subnets that other units will see
    the charm connecting from. Due to things like NAT it isn't always possible to
    narrow it down to a single address, but when it is clear, the CIDRs will
    be constrained to a single address (for example, 10.0.0.1/32).
```

A solution as simple as changing the example docstring in the lib to `str(self.model.get_binding(relation).network.egress_subnets][0].subnet)` could have been enough, however this solution that introduces lists is based on the `Due to things like NAT it isn't always possible to
    narrow it down to a single address, but when it is clear, the CIDRs will
    be constrained to a single address (for example, 10.0.0.1/32)` part of the documentation of the `egress_subnets` 

This change will allow the charm to respect the `--via` option when related in juju, this is important when Vault is related to a kv requirer that is deployed in a different model.

Fixes https://github.com/canonical/vault-operator/issues/244
# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_kv.py
+++ b/lib/charms/vault_k8s/v0/vault_kv.py
@@ -94,7 +94,7 @@ class ExampleRequirerCharm(CharmBase):
         # Update status might not be the best place
         binding = self.model.get_binding("vault-kv")
         if binding is not None:
-            egress_subnet = str(binding.network.interfaces[0].subnet)
+            egress_subnet = [str(subnet) for subnet in self.model.get_binding(relation).network.egress_subnets][0].subnet]
             relation = self.model.get_relation(relation_name="vault-kv")
             self.interface.request_credentials(relation, egress_subnet, self.get_nonce())
 
--- a/lib/charms/vault_k8s/v0/vault_kv.py
+++ b/lib/charms/vault_k8s/v0/vault_kv.py
@@ -219,6 +219,18 @@ class KVRequest:
     nonce: str
 
 
+def get_egress_subnets_list_from_relation_data(relation_databag: Mapping[str, str]) -> List[str]:
+    """Return the egress_subnet as a list.
+
+    This function converts the string with values separated by commas to a list.
+
+    Args:
+        relation_databag: the relation databag of the unit or the app.
+    """
+    print(type(relation_databag))
+    return [subnet.strip() for subnet in relation_databag.get("egress_subnet", "").split(",")]
+
+
 def is_requirer_data_valid(app_data: Mapping[str, str], unit_data: Mapping[str, str]) -> bool:
     """Return whether the requirer data is valid."""
     try:
@@ -339,7 +351,9 @@ class VaultKvProvides(ops.Object):
                 app_name=event.app.name,
                 unit_name=unit.name,
                 mount_suffix=event.relation.data[event.app]["mount_suffix"],
-                egress_subnets=event.relation.data[unit]["egress_subnet"].split(","),
+                egress_subnets=get_egress_subnets_list_from_relation_data(
+                    event.relation.data[unit]
+                ),
                 nonce=event.relation.data[unit]["nonce"],
             )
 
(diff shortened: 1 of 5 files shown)


lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -184,7 +186,7 @@ class AppVaultKvRequirerSchema(BaseModel):
 class UnitVaultKvRequirerSchema(BaseModel):
     """Unit schema of the requirer side of the vault-kv interface."""
 
-    egress_subnet: str = Field(description="Egress subnet to use, in CIDR notation.")
+    egress_subnets: str = Field(description="Egress subnet to use, in CIDR notation.")
```
gruyaume, line 189: This is a change to the interface and should come with a change here as well:
- https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/vault_kv/v0

gruyaume, line 189:
```suggestion
    egress_subnets: str = Field(description="Egress subnets to use, in CIDR notation.")
```

lib/charms/vault_k8s/v0/vault_kv.py
gruyaume, line 1: We haven't bumped the patch version

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: Good point. [PR](https://github.com/canonical/charm-relation-interfaces/pull/163) created.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: libpatch version has been bumped to 10, unless you mean libapi.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: ✅

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -324,10 +339,14 @@ def _on_relation_changed(self, event: ops.RelationChangedEvent):
                 app_name=event.app.name,
                 unit_name=unit.name,
                 mount_suffix=event.relation.data[event.app]["mount_suffix"],
-                egress_subnet=event.relation.data[unit]["egress_subnet"],
+                egress_subnets=event.relation.data[unit]["egress_subnet"].split(","),
```
gruyaume, line 342: Will that work if there's a space after the comma? We should likely get rid of spaces

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: There should never be a space, it's the lib that's puts data into the databag only.
But I fixed it in a new function just in case.

gruyaume: In the change you made to the charm relation interfaces, there's a space after the comma.

https://github.com/canonical/charm-relation-interfaces/pull/163/files

gruyaume (approved):

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: ✅

PR #458, opened 2024-08-15
refactor: use tls library v4

gruyaume:
# Description

Replace the TLS certificates library V3 with V4 for the access integration. 

## Use of self-signed certificates
The Vault charm can effectively operate in two modes for its own tls certificates:
- Self Signed
- TLS integration

Here we explicitly state those two modes. When we use self signed we generate our own certs and when we use the tls integration, we ask certs using the integration.

## This change does not include the PKI integration
In order to make this change a bit smaller we only include here the Access integration. A separate PR will focus on the PKI integration.

# Checklist:


gruyaume:
> **Now that the TLS lib v4 does not provide helpers for generating self signed certs, we have to implement our own helpers. Those are only used when we do not use the TLS integration.**

@ghislainbourgeois I would be interested in your opinion on this. Effectively those helpers would be useful when the tls integration is not used. Do you think we should still include those in the lib.

ghislainbourgeois:
> > **Now that the TLS lib v4 does not provide helpers for generating self signed certs, we have to implement our own helpers. Those are only used when we do not use the TLS integration.**
> 
> @ghislainbourgeois I would be interested in your opinion on this. Effectively those helpers would be useful when the tls integration is not used. Do you think we should still include those in the lib?

I think having them in the lib makes sense, because we use those in different places for different reasons, but those functions are TLS specific. I'd rather have a single place to maintain them.

gruyaume:
> > > **Now that the TLS lib v4 does not provide helpers for generating self signed certs, we have to implement our own helpers. Those are only used when we do not use the TLS integration.**
> > 
> > 
> > @ghislainbourgeois I would be interested in your opinion on this. Effectively those helpers would be useful when the tls integration is not used. Do you think we should still include those in the lib?
> 
> I think having them in the lib makes sense, because we use those in different places for different reasons, but those functions are TLS specific. I'd rather have a single place to maintain them.

All right, I'll add those up:
- https://github.com/canonical/tls-certificates-interface/pull/223

PR #506, opened 2024-09-30
fix: Makes sure there always is an active issuer in Vault PKI

### LINKED ISSUE #500: Vault PKI will stop providing certificates when the CA has less than 90 days to expire
### Bug Description

Vault by default will have 90 days asa validity for the certificates signed using PKI. Vault will also prevent the issuer from signing certificates that would outlive the CA cert of the issuer, therefore the issuer won't provide any certificates when it is 90 days or less to its expiry as the charm currently doesn't renew the issuer before that period.

### To Reproduce

1. Deploy Vault
2. Deploy self signed certs
3. configure self signed certs to use short validity times
4. Integrate both charm
5. Deploy a requirer
6. Integrate the requirer with Vault
7. Observer the issue when the CA has not yet expired but entered the period described above (by default this should happen right away as the charm relies on the default ttl of issued certificates)

### Environment

Not relevant 

### Relevant log output

```shell
cannot satisfy request, as TTL would result in notAfter <date and time> that is beyond the expiration of the CA certificate at <date and time>
```

### Additional context

_No response_

saltiyazan:
# Description

Fixes #500 

- Checks if there is a CA that can issue certificates (Vault does not allow issuing certificates that would outlast the CA)
- Sets the ttl of the pki role to half of the validity of the CA, this will be used by Vault to set validity of issued certificates
- Adds a unit test

**Note to reviewer**: 
- It might help to read spec TE118 while reviewing this.

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_client.py
+++ b/lib/charms/vault_k8s/v0/vault_client.py
@@ -27,7 +27,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 18
+LIBPATCH = 19
 
 
 RAFT_STATE_ENDPOINT = "v1/sys/storage/raft/autopilot/state"
@@ -380,14 +380,27 @@ class Vault:
             logger.warning("Error while signing PKI certificate: %s", e)
             return None
 
-    def create_or_update_pki_charm_role(self, role: str, allowed_domains: str, mount: str) -> None:
-        """Create a role for the PKI backend."""
+    def create_or_update_pki_charm_role(
+        self, role: str, allowed_domains: str, max_ttl: str, mount: str
+    ) -> None:
+        """Create a role for the PKI backend.
+
+        Args:
+            role: The name of the role to create or update.
+            allowed_domains: The list of allowed domains for the role.
+            max_ttl: The maximum TTL for the role.
+                Will be used as validity for the certificates issued by this role.
+                Should be a string in the format of a number with a unit such as
+                "120m", "10h" or "90d".
+            mount: The mount point of the PKI backend for which the role will be created.
+        """
         self._client.secrets.pki.create_or_update_role(
             name=role,
             mount_point=mount,
             extra_params={
                 "allowed_domains": allowed_domains,
                 "allow_subdomains": True,
+                "max_ttl": max_ttl,
             },
         )
         logger.info("Created a role for the PKI backend")
@@ -468,6 +481,18 @@ class Vault:
         except InvalidPath:
             logger.warning("Role does not exist on the specified path.")
             return False
+
+    def get_role_max_ttl(self, role: str, mount: str) -> str:
+        """Get the max ttl for the specified PKI role."""
+        try:
+            return (
+                self._client.secrets.pki.read_role(name=role, mount_point=mount)
+                .get("data", {})
+                .get("max_ttl", [])
+            )
+        except InvalidPath:
+            logger.warning("Role does not exist on the specified path.")
+            return ""
 
     def make_latest_pki_issuer_default(self, mount: str) -> None:
         """Update the issuers config to always make the latest issuer created default issuer."""
--- a/lib/charms/vault_k8s/v0/vault_client.py
+++ b/lib/charms/vault_k8s/v0/vault_client.py
@@ -383,7 +383,7 @@ class Vault:
     def create_or_update_pki_charm_role(
         self, role: str, allowed_domains: str, max_ttl: str, mount: str
     ) -> None:
-        """Create a role for the PKI backend.
+        """Create a role for the PKI backend or update it if it already exists.
 
         Args:
             role: The name of the role to create or update.
@@ -403,7 +403,7 @@ class Vault:
                 "max_ttl": max_ttl,
             },
         )
-        logger.info("Created a role for the PKI backend")
+        logger.info("Created or updated PKI role %s", role)
 
     def is_pki_role_created(self, role: str, mount: str) -> bool:
         """Check if the role is created for the PKI backend."""
@@ -488,7 +488,7 @@ class Vault:
             return (
                 self._client.secrets.pki.read_role(name=role, mount_point=mount)
                 .get("data", {})
-                .get("max_ttl", [])
+                .get("max_ttl")
             )
         except InvalidPath:
             logger.warning("Role does not exist on the specified path.")
--- a/lib/charms/vault_k8s/v0/vault_client.py
+++ b/lib/charms/vault_k8s/v0/vault_client.py
@@ -352,6 +352,7 @@ class Vault:
         role: str,
         csr: str,
         common_name: str,
+        ttl: str,
     ) -> Certificate | None:
         """Sign a certificate signing request for the PKI backend.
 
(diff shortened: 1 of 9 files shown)


lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -469,6 +482,18 @@ def is_common_name_allowed_in_pki_role(self, role: str, mount: str, common_name:
             logger.warning("Role does not exist on the specified path.")
             return False
 
+    def get_role_max_ttl(self, role: str, mount: str) -> Optional[int]:
```
gruyaume, line 485: In the method above, the max_ttl was passed in as a string (ex. "120m"), when we run a get now we get an int (?). Is that really the case and if yes, what does it represent? hours?

gruyaume: Reading charm.py it seems like it is seconds. Let's make sure that's correct and let's document this in the method docstrings.

src/charm.py
```diff
@@ -537,22 +547,62 @@ def _configure_pki_secrets_engine(self) -> None:  # noqa: C901
...
+            - intermediate_ca_certificate.validity_start_time
+        )
+        certificate_validity_seconds = certificate_validity.total_seconds()
+        return certificate_validity_seconds > current_ttl
+
+    def _calculate_pki_certificates_validity(self, certificate: Certificate) -> int:
+        """Calculate the maximum allowed validity of certificates issued by PKI.
+
+        Return half the CA certificate validity.
+        """
```
gruyaume, line 599: What's the unit here? It seems to be seconds, but let's make that explicit

src/charm.py
```diff
@@ -537,22 +547,62 @@ def _configure_pki_secrets_engine(self) -> None:  # noqa: C901
...
+        ):
+            return False
+        certificate_validity = (
+            intermediate_ca_certificate.expiry_time
+            - intermediate_ca_certificate.validity_start_time
+        )
+        certificate_validity_seconds = certificate_validity.total_seconds()
+        return certificate_validity_seconds > current_ttl
+
+    def _calculate_pki_certificates_validity(self, certificate: Certificate) -> int:
```
gruyaume, line 595: I'd suggest a better function name, validities are date typically, here it's more of a ttl / time delta.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -380,17 +380,30 @@ def sign_pki_certificate_signing_request(
...
+                "120m", "10h" or "90d".
+            mount: The mount point of the PKI backend for which the role will be created.
+        """
         self._client.secrets.pki.create_or_update_role(
             name=role,
             mount_point=mount,
             extra_params={
                 "allowed_domains": allowed_domains,
                 "allow_subdomains": True,
+                "max_ttl": max_ttl,
```
gruyaume, line 403: Here we set the max_ttl, what's the actual ttl that's being used when signing certificates, is it the same value?

src/charm.py
```diff
@@ -537,22 +547,62 @@ def _configure_pki_secrets_engine(self) -> None:  # noqa: C901
...
+        """Calculate the maximum allowed validity of certificates issued by PKI.
+
+        Return half the CA certificate validity.
```
gruyaume, line 598: We should likely document this in our documentation (after this is merged). It should be quite clear that "when using Vault as an intermediate CA authority, its own certificate validity will be determined by the TLS providers and the certificates it issues will be valid for half this time."

src/charm.py
```diff
@@ -537,22 +547,62 @@ def _configure_pki_secrets_engine(self) -> None:  # noqa: C901
...
+            or not intermediate_ca_certificate.expiry_time
+            or not intermediate_ca_certificate.validity_start_time
+        ):
+            return False
+        certificate_validity = (
+            intermediate_ca_certificate.expiry_time
+            - intermediate_ca_certificate.validity_start_time
+        )
+        certificate_validity_seconds = certificate_validity.total_seconds()
+        return certificate_validity_seconds > current_ttl
```
gruyaume, line 593: It feels like the method name does not match with its content. Being active to me sounds closer to "is the pki secret engine enabled".

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_client.py
saltiyazan: Yes, this is a weird one, but that's what the API returns.
Docstring updated

src/charm.py
saltiyazan: Docstring updated

src/charm.py
saltiyazan: Changed to `_calculate_pki_certificates_validity_period`

lib/charms/vault_k8s/v0/vault_client.py
saltiyazan: We sign certificates using `client.secrets.pki.sign_certificate` that could take a `ttl` in its `extra_params` and that would be the validity of the signed certificate. However this can't exceed the `max_ttl` that is set for the role of the issuer. Additionally if no `ttl` is specified in `client.secrets.pki.sign_certificate` then the `max_ttl` of the role will be used by default.

I decided to manage this centrally when the PKI engine is being configured but I improved the documentation to make sure this is clear.

If you prefer to increase readability by setting this in the `client.secrets.pki.sign_certificate` too or by moving it completely there I'd be happy to do that.

src/charm.py
saltiyazan: Yes, I agree.

src/charm.py
saltiyazan: Changed to `_intermediate_ca_can_sign_certificates`

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: I think it would be good from a developer point of view to also set the ttl in the `sign_certificate` method, this way we know for sure what's the cert validity. Right now we have to infer and guess

src/charm.py
```diff
@@ -537,22 +542,62 @@ def _configure_pki_secrets_engine(self) -> None:  # noqa: C901
...
                 role=PKI_ROLE_NAME,
+                max_ttl=f"{issued_certificates_validity}s",
             )
         # Can run only after the first issuer has been actually created.
         try:
             vault.make_latest_pki_issuer_default(mount=PKI_MOUNT)
         except VaultClientError as e:
             logger.error("Failed to make latest issuer default: %s", e)
 
+    def _intermediate_ca_can_sign_certificates(
```
gruyaume, line 567: I'd recommend going with exactly what this method does: `intermediate_ca_exceeds_role_ttl`

src/charm.py
```diff
@@ -537,22 +542,62 @@ def _configure_pki_secrets_engine(self) -> None:  # noqa: C901
...
+        ):
+            return False
+        certificate_validity = (
+            intermediate_ca_certificate.expiry_time
+            - intermediate_ca_certificate.validity_start_time
+        )
+        certificate_validity_seconds = certificate_validity.total_seconds()
+        return certificate_validity_seconds > current_ttl
+
+    def _calculate_pki_certificates_validity_period(self, certificate: Certificate) -> int:
```
gruyaume, line 590:
```suggestion
    def _calculate_pki_certificates_ttl(self, certificate: Certificate) -> int:
```

gruyaume (changes requested):
Looks mostly good, a couple of nits related to method names. We have many many methods in this file and I want to make sure their naming is precise.

src/charm.py
saltiyazan: ✅

src/charm.py
saltiyazan: ✅

lib/charms/vault_k8s/v0/vault_client.py
saltiyazan: Done

gruyaume (approved):

PR #509, opened 2024-10-03
chore: Upload charm to latest/edge when main branch is updated

saltiyazan:
# Description

Currently we only release the charm to the release tracks (`1.15` and `1.16`).
In this change we add a workflow to release to `latest/edge. This track will follow main which means it will have the same revision as in `1.16/edge` for now and later when there is a new release it will follow that release (for example `1.17`)

# Checklist:


gruyaume (changes requested):
Why release to `latest` at all? We should only release to release tracks (ex. `1.15`, `1.16`, `1.17`).

PR #531, opened 2024-11-14
feat: Implement the JujuFacade wrapper class

saltiyazan:
# Description

Implements the `JujuFacade` class and uses it in `VaultTLSManager`
This PR implements TE109.

It regroups exceptions and raises them wrapped in exceptions defined in the library.

For example `get_secret` would still raise an exception if the secret was not found, wrapped in `NoSuchSecretError` however the advantage of juju_facade is that the caller doesn't need to run `get_secret`, then `secret.get_content` and finally `set_content` to change the secret content. It is all wrapped into one call. For more info please read the spec.

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/juju_facade.py
+++ b/lib/charms/vault_k8s/v0/juju_facade.py
@@ -76,7 +76,12 @@ class JujuFacade(Object):
 
     # Secret related methods
     def get_secret(self, label: str, id: str | None = None) -> Secret:
-        """Retrieve a secret and handle SecretNotFound error."""
+        """Retrieve a secret and handles error.
+
+        Raises:
+            NoSuchSecretError: if the secret does not exist
+            TransientJujuError: if there is a Juju ModelError
+        """
         try:
             return (
                 self.charm.model.get_secret(label=label, id=id)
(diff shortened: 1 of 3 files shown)


lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -0,0 +1,522 @@
...
+        relation_name: str,
+        relation_id: int,
+        entity: Unit | Application,
+    ) -> None:
+        relation = self.get_relation_by_id(relation_name, relation_id)
+        if not relation:
+            raise NoSuchRelationError(f"Relation {relation_name}:{relation_id} not found")
+        if not all(isinstance(value, str) for value in data.values()) or not all(
+            isinstance(key, str) for key in data.keys()
+        ):
```
ghislainbourgeois, line 426: This could be a bit less verbose by using `chain()` from `itertools`:

```suggestion
from itertools import chain
...

        if not all(isinstance(value, str) for value in chain(data.values(), data.keys())):
```

This is just suggestion, I would not block the PR for this.

ghislainbourgeois (approved):

kayra1 (commented):
This is a really good start, I think it may be nice to have @DanielArndt 's opinion on it too.

PR #534, opened 2024-11-20
Vault client refactor

DanielArndt:
# Description

Please include a summary of the change. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Checklist:


tox.ini
```diff
@@ -16,7 +16,7 @@ all_path = {[vars]src_path} {[vars]unit_test_path} {[vars]integration_test_path}
 
 [testenv]
 set_env =
-    PYTHONPATH = {toxinidir}:{toxinidir}/lib:{[vars]src_path}
+    PYTHONPATH = {toxinidir}:{toxinidir}/lib:{[vars]src_path}:{[vars]integration_test_path}/vault_kv_requirer_operator/src
```
DanielArndt, line 19: This gets rid of the need for a `# type: ignore` in the test code that masked one of my errors -- alternative would be to break up the `pyright` calls and only provide this path when type checking the specific file.

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -515,68 +576,345 @@ def make_latest_pki_issuer_default(self, mount: str) -> None:
...
+        existing_token = self._get_juju_secret_field(self.AUTOUNSEAL_TOKEN_SECRET_LABEL, "token")
+        # If we don't already have a token, or if the existing token is invalid,
+        # authenticate with the AppRole details to generate a new token.
+        if not existing_token or not external_vault.authenticate(Token(existing_token)):
+            external_vault.authenticate(
+                AppRole(autounseal_details.role_id, autounseal_details.secret_id)
+            )
+            # NOTE: This is a little hacky. If the token expires, every unit
+            # will generate a new token, until the leader unit generates a new
+            # valid token and sets it in the Juju secret.
```
DanielArndt, line 883: It might be better to just put the charm in waiting status here...

DanielArndt:
There is an [intermittent integration test failure](https://github.com/canonical/vault-k8s-operator/actions/runs/11958740844/job/33339786512), but I believe it is unrelated to this change (although it may be exacerbated by it).

I'm looking into it.

lib/charms/vault_k8s/v0/vault_client.py
gruyaume: We probably don't want to have the charm's status management handled by the Vault API client

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -169,6 +177,43 @@ def is_sealed(self) -> bool:
...
+    def read(self, path: str) -> dict:
+        """Read the data at the given path."""
+        try:
+            data = self._client.read(path)
+        except VaultError as e:
+            logger.error("Error while writing data to %s: %s", path, e)
+            return {}
+        if data is None:
+            return {}
+        assert isinstance(data, dict)
```
gruyaume, line 189: Do we really want those asserts in the code and end up with potential assertion errors?

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -515,68 +576,345 @@ def make_latest_pki_issuer_default(self, mount: str) -> None:
...
+
+
+AUTOUNSEAL_POLICY = """path "{mount}/encrypt/{key_name}" {{
+    capabilities = ["update"]
+}}
+
+path "{mount}/decrypt/{key_name}" {{
+    capabilities = ["update"]
+}}
+"""
```
gruyaume, line 605: Should this live here or in the autounseal lib?

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -515,68 +576,345 @@ def make_latest_pki_issuer_default(self, mount: str) -> None:
         except (TypeError, KeyError):
             logger.error("Issuers config is not yet created")
 
-    def _get_autounseal_policy_name(self, relation_id: int) -> str:
-        """Return the policy name for the given relation id."""
-        return f"charm-autounseal-{relation_id}"
+    def create_transit_key(self, mount_point: str, key_name: str) -> None:
+        """Create a new key in the transit backend."""
+        response = self._client.secrets.transit.create_key(mount_point=mount_point, name=key_name)
+        logging.debug(f"Created a new transit key. response={response}")
```
gruyaume, line 582: Let's avoid f-strings in logs.

Reference:
- https://docs.google.com/document/d/1H5W2oi7cuTaz86taTQiITv49jpJNKzzDZ2G25IQpE4w/edit?tab=t.0#heading=h.84ez402ai5cq

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -515,68 +576,345 @@ def make_latest_pki_issuer_default(self, mount: str) -> None:
...
+}}
+
+path "{mount}/decrypt/{key_name}" {{
+    capabilities = ["update"]
+}}
+"""
+
+
+@dataclass
+class VaultAutounsealRelationDetails:
```
gruyaume, line 609: I'm somewhat confused by this file now, it used to be quite clear what it was (a Vault http api abstraction), now it's a stranger, larger thing that does a lot. 

Does this class really belong in `vault_client.py`?

gruyaume: The same comment may apply to other classes here

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: @gruyaume Yeah, it would have to be done via a return value.

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: I guess this is a matter of preference. My preference is yes.

I threw those in because hvac doesn't have any type hinting. I'm assuming they won't change, but I don't know. The options as I see it are:

- We could add a `try..except` block, but we still want it to fail if this ever changes in a newer version of the lib. So the effect would be the same.
- We could add a type ignore, which I am strongly averse to.
- By force of habit, I use asserts for assumptions like this because they can always be optimized out.

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: In short, probably not, we will discuss face-to-face.

DanielArndt: Moved in https://github.com/canonical/vault-k8s-operator/pull/534/commits/a09aefe4129ccad0ad30965b0de2b1320de6a77a

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: Fixed in https://github.com/canonical/vault-k8s-operator/commit/a09aefe4129ccad0ad30965b0de2b1320de6a77a

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
+"""Library for managing Vault Charm features.
+
+This library encapsulates the business logic for managing the Vault service and
+its associated integrations within the context of our charms.
+"""
+
+# The unique Charmhub library identifier, never change it
```
gruyaume, line 7: This comment should be next to the lib id?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+
+    @property
+    def key_name(self) -> str:
+        """Return the key name for the relation."""
+        return str(self.relation.id)
+
+    @property
+    def policy_name(self) -> str:
+        """Return the policy name for the relation."""
+        return f"charm-autounseal-{self.relation.id}"
```
gruyaume, line 82: This prefix shows up a lot in this file, should we use a variable?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        mount_path: str = "charm-autounseal",
+    ):
+        self._charm = charm
+        self._juju_facade = JujuFacade(charm)
+        self._model = juju_model
+        self._client = client
+        self._provides = provides
+        self._mount_path = mount_path
+        self._ca_cert = ca_cert
+        self._port = 8200
```
gruyaume, line 115: This port is defined both here and in `charm.py`

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+
+    In other words, this manages the feature from the perspective of the Vault
+    being auto-unsealed.
+    """
+
+    AUTOUNSEAL_TOKEN_SECRET_LABEL = "vault-autounseal-token"
+
+    def __init__(
+        self,
+        tls_manager: VaultTLSManager,
```
gruyaume, line 285: We probably don't want managers to depend on each others right?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        existing_keys = self._get_existing_keys()
+        relation_key_names = [
+            VaultAutounsealRelationDetails(relation).key_name
+            for relation in self._juju_facade.get_active_relations(self._provides.relation_name)
+        ]
+        orphaned_keys = [key for key in existing_keys if key not in relation_key_names]
+        if not orphaned_keys:
+            return
+        logging.warning(
+            f"Orphaned autounseal keys were detected: {orphaned_keys}. If you are sure these are no longer needed, you may manually delete them using the vault CLI to suppress this message."
```
gruyaume, line 174: Let's avoid f-strings in log messages

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+            " To delete a key, use the command `vault delete charm-autounseal/keys/<key_name>`."
+        )
+        for key_name in orphaned_keys:
+            deletion_allowed = self._detect_if_deletion_allowed(key_name)
+            if not deletion_allowed:
+                self._allow_key_deletion(key_name)
+
+    def _allow_key_deletion(self, key_name) -> None:
```
gruyaume, line 182: We're missing the input type

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        logger.info("Allowing deletion of key %s", key_name)
+        self._client.write(f"{self.mount_path}/keys/{key_name}/config", {"deletion_allowed": True})
+
+    def _detect_if_deletion_allowed(self, key_name) -> bool:
```
gruyaume, line 186: We're missing the input type

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        relation_policy_names = [
+            VaultAutounsealRelationDetails(relation).policy_name
+            for relation in self._juju_facade.get_active_relations(self._provides.relation_name)
+        ]
+        for policy in existing_policies:
+            if policy not in relation_policy_names:
+                logging.info("Removing unused policy: %s", policy)
+                self._client.delete_policy(policy)
+
+    def _create_key(self, key_name) -> None:
```
gruyaume, line 212: We're missing the input type

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+
+        This method will remove any roles and policies that are no longer
+        used by any of the existing relations. It will also detect any orphaned
+        keys (keys that are not associated with any relation) and log a warning.
+        """
+        self._clean_up_roles()
+        self._clean_up_policies()
+        self._detect_orphaned_keys()
+
+    def _detect_orphaned_keys(self) -> None:
```
gruyaume, line 164: This method does more than "detecting", it also allows their deletion

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+            " To delete a key, use the command `vault delete charm-autounseal/keys/<key_name>`."
+        )
+        for key_name in orphaned_keys:
+            deletion_allowed = self._detect_if_deletion_allowed(key_name)
+            if not deletion_allowed:
+                self._allow_key_deletion(key_name)
+
+    def _allow_key_deletion(self, key_name) -> None:
+        logger.info("Allowing deletion of key %s", key_name)
+        self._client.write(f"{self.mount_path}/keys/{key_name}/config", {"deletion_allowed": True})
```
gruyaume, line 184: Let's move the log after the action was completed, just like we have elsewhere here.

```
self._client.write(f"{self.mount_path}/keys/{key_name}/config", {"deletion_allowed": True})
logger.info("Allowed deletion of key %s", key_name)
```

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        self,
+        tls_manager: VaultTLSManager,
+        model: Model,
+        requires: VaultAutounsealRequires,
+    ):
+        self._tls_manager = tls_manager
+        self._model = model
+        self._requires = requires
+
+    def vault_configuration_details(self) -> AutounsealConfigurationDetails | None:
```
gruyaume, line 293: should we have a `get` prefix?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        """Return the configuration details for the vault."""
```
gruyaume, line 294: "configuration default for the vault" is very generic

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        try:
+            juju_secret = self._model.get_secret(label=label)
+        except SecretNotFoundError:
+            return None
+        content = juju_secret.get_content(refresh=True)
+        return content.get(field)
+
+    def _set_juju_secret(
+        self, label: str, content: dict[str, str], description: str | None = None
```
gruyaume, line 359: Why the "description" field that is never used?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+            If the secret does not exist, `None` is returned.
+        """
+        try:
+            juju_secret = self._model.get_secret(label=label)
+        except SecretNotFoundError:
+            return None
+        content = juju_secret.get_content(refresh=True)
+        return content.get(field)
+
+    def _set_juju_secret(
```
gruyaume, line 358: Shouldn't this method belong in Juju facade?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+            # NOTE: This is a little hacky. If the token expires, every unit
+            # will generate a new token, until the leader unit generates a new
+            # valid token and sets it in the Juju secret.
+            if self._model.unit.is_leader():
+                self._set_juju_secret(
+                    self.AUTOUNSEAL_TOKEN_SECRET_LABEL, {"token": external_vault.token}
+                )
+        return external_vault.token
+
+    def _get_juju_secret_field(self, label: str, field: str) -> str | None:
```
gruyaume, line 342: Shouldn't this method belong in Juju facade?

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Yes, sorry. This was written before juju_facade, but I should be able to get rid of these now.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: poor copy from elsewhere, forgot to revisit this.

In any case, it will be removed now that we have juju_facade

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -266,7 +303,23 @@ def configure_policy(self, policy_name: str, policy_path: str, **formatting_args
             raise VaultClientError(e) from e
         logger.debug("Created or updated charm policy: %s", policy_name)
 
-    def configure_approle(
+    def create_or_update_policy(self, policy_name: str, policy_content: str) -> None:
```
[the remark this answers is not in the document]
DanielArndt: Hmm, there might have been a bit of a merge mistake here, too, as this was supposed to be a simple rename of the function above.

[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/c6cdfab45a786e77d34c0db7f035d8a269a7d815

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: :D -- auto imports :)

oops

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/1ab346f30cc700a7d01eee6d821e2757552fc977

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/1ab346f30cc700a7d01eee6d821e2757552fc977

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/1ab346f30cc700a7d01eee6d821e2757552fc977

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/1ab346f30cc700a7d01eee6d821e2757552fc977

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -515,66 +568,18 @@ def make_latest_pki_issuer_default(self, mount: str) -> None:
...
-        self, relation_id: int, mount: str, policy_path: str
-    ) -> tuple[str, str, str]:
-        """Create auto-unseal credentials for the given relation id.
+        logging.debug("Created a new transit key. response=%s", response)
 
-        Args:
-            relation_id: The Juju relation id to use for the approle.
-            mount: The mount point for the transit backend.
-            policy_path: Path to a file that contains the autounseal policy.
+    def delete_role(self, role_name: str) -> None:
```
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/2e06506d24b44f6c3a515cfbe12fa29c088b9d41

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
+"""Library for managing Vault Charm features.
+
+This library encapsulates the business logic for managing the Vault service and
+its associated integrations within the context of our charms.
+"""
```
gruyaume, line 5: Can we add a bit more content here. It would be great to have:
- The definition of a manager
- What belongs here, and what doesn't
- Any design principle as to what a manager should look like (Ex. with the `sync` approach)

lib/charms/vault_k8s/v0/vault_client.py
```diff
@@ -307,7 +344,7 @@ def read_role_secret(self, name: str, id: str) -> dict:
         response = self._client.auth.approle.read_secret_id(name, id)
         return response["data"]
 
-    def enable_secrets_engine(self, backend_type: SecretsBackend, path: str) -> None:
+    def ensure_secrets_engine(self, backend_type: SecretsBackend, path: str) -> None:
```
gruyaume, line 347: Why the name change? I'm not convinced `ensure` is better

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
...
+        charm: CharmBase,
+        juju_model: Model,
+        client: VaultClient,
+        provides: VaultAutounsealProvides,
+        ca_cert: str,
+        mount_path: str = "charm-autounseal",
+    ):
+        self._charm = charm
+        self._juju_facade = JujuFacade(charm)
+        self._model = juju_model
```
gruyaume, line 110: We can probably only require JujuFacade (and remove the `model` input parameter) if we add the necessary methods in JujuFacade

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
...
+    def __init__(
+        self,
+        charm: CharmBase,
+        juju_model: Model,
+        client: VaultClient,
+        provides: VaultAutounsealProvides,
+        ca_cert: str,
+        mount_path: str = "charm-autounseal",
+    ):
+        self._charm = charm
```
gruyaume, line 108: Where is this used?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
...
+        self._juju_facade = JujuFacade(charm)
+        self._model = juju_model
+        self._client = client
+        self._provides = provides
+        self._mount_path = mount_path
+        self._ca_cert = ca_cert
+        self._port = 8200
+
+        if not self._ca_cert:
+            logger.warning("No CA certificate provided for auto-unseal")
```
gruyaume, line 118: Is this possible as Vault charms are always deployed with TLS? If we go with the assertion approach, shouldn't we simply assert that the ca certificate was provided.

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: nowhere, removed in https://github.com/canonical/vault-k8s-operator/pull/534/commits/0fbf37fae0e6926b5dad35cf8817fab997dd5529

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: fixed in https://github.com/canonical/vault-k8s-operator/pull/534/commits/0fbf37fae0e6926b5dad35cf8817fab997dd5529

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: ensure implies (to me) that it is idempotent, whereas enable doesn't. I was trying to address the same issue we had with `configure_*`, and we seemed to be converging on `create_or_update_` for those.

I can change it back if it's disliked.

gruyaume: `create_or_update_` won't create or update anything if the field is already in its correct value right? Ensure is too vague in my opinion, ensure what? That it is enabled, disabled, created, etc. I do agree though on having every method be idepotent, but I don't think it should mean renaming all of them to `ensure_xyz`

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,372 @@
...
+    VaultAutounsealRequires,
+)
+from charms.vault_k8s.v0.vault_client import (
+    AppRole,
+    SecretsBackend,
+    Token,
+    VaultClient,
+    VaultClientError,
+)
+from charms.vault_k8s.v0.vault_tls import File, VaultTLSManager
```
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/8c7afaaec02342197b7d7d03e3b353ee8d1c99b4

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -545,6 +549,15 @@ def get_storage_location(self, storage_name: str) -> Path:
             raise NoSuchStorageError(f"Storage {storage_name} not found")
         return storages[storage_name][0].location
 
+    def get_ingress_address(self, relation) -> str | None:
```
gruyaume, line 552: We're missing the input type

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -545,6 +549,15 @@ def get_storage_location(self, storage_name: str) -> Path:
...
+        """Get the ingress address."""
```
gruyaume, line 553: Is this an IP address? The specific format would be useful here, ex. "1.2.3.4", "http://my_address.com", something else.

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +526,288 @@ def generate_vault_unit_certificate(
...
+        for key_name in orphaned_keys:
+            deletion_allowed = self._is_deletion_allowed(key_name)
+            if not deletion_allowed:
+                self._allow_key_deletion(key_name)
+
+    def _allow_key_deletion(self, key_name) -> None:
+        logger.info("Allowing deletion of key %s", key_name)
+        self._client.write(f"{self.mount_path}/keys/{key_name}/config", {"deletion_allowed": True})
+
+    def _is_deletion_allowed(self, key_name) -> bool:
```
gruyaume, line 652: We're missing the input type

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
...
+
+        if not self._ca_cert:
+            logger.warning("No CA certificate provided for auto-unseal")
+
+    def get_address(self, relation: Relation) -> str:
+        """Fetch the address from the relation and return it."""
+        binding = self._model.get_binding(relation)
+        if binding is None:
+            raise VaultClientError("Failed to fetch binding from relation")
+        return f"https://{binding.network.ingress_address}:{self._port}"
```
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/6d7363b223fcad19b9788bde3928459078149034

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
...
+
+        This looks for any outstanding requests for auto-unseal that may have
+        been missed. If there are any, it generates the credentials and sets
+        them in the relation databag.
+
+        It also cleans up any credentials that are no longer used by any of the
+        relations, and logs a warning about orphaned keys. It will not remove
+        any keys, to prevent loss of data.
+        """
+        if not self._model.unit.is_leader():
```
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/6d7363b223fcad19b9788bde3928459078149034

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +526,288 @@ def generate_vault_unit_certificate(
...
+
+    def _allow_key_deletion(self, key_name) -> None:
+        logger.info("Allowing deletion of key %s", key_name)
+        self._client.write(f"{self.mount_path}/keys/{key_name}/config", {"deletion_allowed": True})
+
+    def _is_deletion_allowed(self, key_name) -> bool:
+        data = self._client.read(f"{self.mount_path}/keys/{key_name}")
+        return data["deletion_allowed"]
+
+    def _clean_up_roles(self) -> None:
```
gruyaume, line 656: I feel docstrings would be useful here to understand what roles we are deleting exactly and why. The same comment likely applies to policies. Example `Delete autounseal Vault roles that are not associated with an existing autounseal Juju integration`

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +526,288 @@ def generate_vault_unit_certificate(
...
+
+    AUTOUNSEAL_TOKEN_SECRET_LABEL = "vault-autounseal-token"
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        tls_manager: VaultTLSManager,
+        requires: VaultAutounsealRequires,
+    ):
+        self._charm = charm
```
gruyaume, line 755: This is not used anymore

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +526,288 @@ def generate_vault_unit_certificate(
...
+
+    In other words, this manages the feature from the perspective of the Vault
+    being auto-unsealed.
+    """
+
+    AUTOUNSEAL_TOKEN_SECRET_LABEL = "vault-autounseal-token"
+
+    def __init__(
+        self,
+        charm: CharmBase,
```
gruyaume, line 751: This is not used anymore

src/charm.py
gruyaume, line 1: This file looks much cleaner now!

src/container.py
```diff
@@ -6,7 +6,7 @@
 
 from typing import TextIO
 
-from charms.vault_k8s.v0.vault_tls import WorkloadBase
+from charms.vault_k8s.v0.vault_managers import WorkloadBase
```
gruyaume, line 9: Why do we import this from the vault managers? It feels strange for container.py to depend on Vault managers (or on Vault TLS for that matter).

I understand this is not quite part of this change here so feel free to ignore my comment.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: This is still used by the `juju_facade`, but you're right we can get rid of the instance variable.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/d073a60bd344366743b70b034ae17d6aae57bebf

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/ab549ee6c4957d3caa79a70f47db804a4e9c142b

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/ab549ee6c4957d3caa79a70f47db804a4e9c142b

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/ab549ee6c4957d3caa79a70f47db804a4e9c142b

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/dd8dc148fb8c47c4890639d6ee8e4f72856581cd

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/dd8dc148fb8c47c4890639d6ee8e4f72856581cd

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def _clean_up_policies(self) -> None:
+        existing_policies = self._get_existing_policies()
+        relation_policy_names = [
+            VaultAutounsealRelationDetails(relation).policy_name
+            for relation in self._juju_facade.get_active_relations(self._provides.relation_name)
+        ]
+        for policy in existing_policies:
+            if policy not in relation_policy_names:
+                logging.info("Removing unused policy: %s", policy)
```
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/a7bdcbb7669d46a06eb1cf2dc01ada7232fe6f9a

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/53f215012786fc13fb764915c729fee1a0c047ef

lib/charms/vault_k8s/v0/juju_facade.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/6e750f519fc9a3f086e72f8c72290465fd010926

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/ab549ee6c4957d3caa79a70f47db804a4e9c142b

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/f27f121018cf8949cc3e1db4a5a6c490bb990907

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: There's a couple different cases, but you're right each one is used in a couple of other places.

I'm not sure this class serves its original purpose anymore, so I will revisit it.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Following our discussion, I'll move this to the charm, but I'm going to save it until near the end because it will be a more functional shift, and I'd like to be more careful with it.

lib/charms/vault_k8s/v0/vault_client.py
DanielArndt: If you have any preference, let me know. I left this comment to spark discussion if there is any preference here, but I went with the easiest option.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: I took a stab at this here: https://github.com/canonical/vault-k8s-operator/pull/534/commits/4adb4c949ae9806da229804583c7eb83efdbab0d

Let me know your thoughts. I don't love the name.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/8a2350e3313db1a915d845bdac3ef778d86e4ec4

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Let me know what you think: https://github.com/canonical/vault-k8s-operator/pull/534/commits/7454c11fd16a2d5951808dd205ff143f266cee2e

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +550,304 @@ def generate_vault_unit_certificate(
...
+    """
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        client: VaultClient,
+        provides: VaultAutounsealProvides,
+        ca_cert: str,
+        vault_port: int,
+        mount_path: str = "charm-autounseal",
```
gruyaume, line 602: Do we need to have a default value here?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +550,304 @@ def generate_vault_unit_certificate(
...
+        self._model = charm.model
+        self._client = client
+        self._provides = provides
+        self._mount_path = mount_path
+        self._ca_cert = ca_cert
+        self._port = vault_port
+
+    def get_address(self, relation: Relation) -> str:
+        """Fetch the address from the relation and return it."""
+        return f"https://{self._juju_facade.get_ingress_address(relation)}:{self._port}"
```
gruyaume, line 614: It doesn't feel like this getter belongs in this class.

Here we assume that the vault service is available at this ingress address but this behavior (address selection) is not handled by autounseal. Shouldn't we pass in the Vault address as an input parameter (instead of the port)?

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
```diff
@@ -0,0 +1,142 @@
...
+
+        key_name, role_id, secret_id = autounseal.create_credentials(relation)
+
+        assert key_name == str(relation_id)
+        assert role_id == "role_id"
+        assert secret_id == "secret_id"
+        provides.set_autounseal_data.assert_called_once()
+
+    @patch("charms.vault_k8s.v0.vault_managers.JujuFacade")
+    def test_sync(self, juju_facade_mock):
```
gruyaume, line 96: Let's follow the test naming convention we have `given_x_when_sync_then_z`

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
```diff
@@ -0,0 +1,142 @@
...
+        assert details is not None
+        assert details.address == "my_address"
+        assert details.mount_path == "my_mount_path"
+        assert details.key_name == "my_key_name"
+        assert details.token == expected_token
+        assert details.ca_cert_path == "/my/test/path"
+
+
+class TestVaultAutounsealProviderManager:
+    def test_create_credentials(self):
```
gruyaume, line 76: Let's follow the test naming convention we have

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
```diff
@@ -0,0 +1,142 @@
...
+        provides.set_autounseal_data.assert_called()
+        vault_client_mock.delete_role.assert_called_once_with("charm-autounseal-321")
+        vault_client_mock.delete_policy.assert_called_once_with("charm-autounseal-321")
+        vault_client_mock.write.assert_called_with(
+            "charm-autounseal/keys/charm-autounseal-321/config", {"deletion_allowed": True}
+        )
+
+    @patch("charms.vault_k8s.v0.vault_client.VaultClient")
+    @patch("charms.vault_k8s.v0.vault_managers.JujuFacade")
+    def test_get_address(self, juju_facade_mock, vault_client_mock):
```
gruyaume, line 130: Let's follow the test naming convention we have

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_managers.py
gruyaume: I think it's great, thank you!

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +550,304 @@ def generate_vault_unit_certificate(
 def existing_certificate_is_self_signed(ca_certificate: Certificate) -> bool:
     """Return whether the certificate is a self signed certificate generated by the Vault charm."""
     return ca_certificate.common_name == VAULT_CA_SUBJECT
+
+
+class VaultAutounsealNaming:
```
gruyaume, line 555: We'll need such names for other features right? Ex for kv we may need a `charm-kv-<relation id>` policy (or something like this). Should we remove the Autounseal from the class name and allow for other features to use it as well?

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +550,304 @@ def generate_vault_unit_certificate(
...
+    This class provides a central place to manage the naming conventions for
+    the auto-unseal functionality.
+    """
+
+    key_prefix: str = ""
+    policy_prefix: str = "charm-autounseal-"
+    approle_prefix: str = "charm-autounseal-"
+
+    @classmethod
+    def key_name(cls, relation_id) -> str:
```
gruyaume, line 571: We're missing the input type

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +550,304 @@ def generate_vault_unit_certificate(
...
+        """Return the key name for the relation."""
+        return f"{cls.key_prefix}{relation_id}"
+
+    @classmethod
+    def policy_name(cls, relation_id) -> str:
```
gruyaume, line 576: We're missing the input type

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -504,3 +550,304 @@ def generate_vault_unit_certificate(
...
+        """Return the policy name for the relation."""
+        return f"{cls.policy_prefix}{relation_id}"
+
+    @classmethod
+    def approle_name(cls, relation_id) -> str:
```
gruyaume, line 581: We're missing the input type

(Does one of our tool have an automatic checker for this?)

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/c35ab15147e33882eac1d1a8e6ffd3a7a694ac6e

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/c35ab15147e33882eac1d1a8e6ffd3a7a694ac6e

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/296e23fbc6af92cfe1d3aaabc0cf99bfb08cd2ab

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
DanielArndt: Removed

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
DanielArndt: Addressed in https://github.com/canonical/vault-k8s-operator/pull/534/commits/b73344ca50ae70673a93f4552c470acdd2fd54f4

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
DanielArndt: Removed in https://github.com/canonical/vault-k8s-operator/pull/534/commits/b73344ca50ae70673a93f4552c470acdd2fd54f4

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Sure, we can move this back to the charm, but it pushes a little more logic back there as well, because the address depends on the relation. We can no longer rely on a single `sync()` call.

https://github.com/canonical/vault-k8s-operator/pull/534/commits/b73344ca50ae70673a93f4552c470acdd2fd54f4

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Addressed in https://github.com/canonical/vault-k8s-operator/pull/534/commits/b73344ca50ae70673a93f4552c470acdd2fd54f4

lib/charms/vault_k8s/v0/juju_facade.py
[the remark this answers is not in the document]
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/f2e92bff6fc5d478793d396346f441d76fa429e9

gruyaume, line 1: FYI this PR conflicts with #539

src/charm.py
```diff
@@ -433,7 +332,7 @@ def _configure(self, _: EventBase) -> None:  # noqa: C901
         if not (vault := self._get_active_vault_client()):
             return
         self._configure_pki_secrets_engine()
-        self._sync_vault_autounseal()
+        self._sync_vault_autounseal(vault)
```
gruyaume, line 335: Why does `_sync_vault_autounseal` take the vault client as an argument and the other syncs (ex. kv) get their vault client inside on their own?

gruyaume (changes requested):
One last comment related to the way we get and move the vault client. Asides from that one I'm ready to approve the PR

lib/charms/vault_k8s/v0/juju_facade.py
DanielArndt: Synced with @saltiyazan -- we'll just deal with conflicts when they arise.

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -26,7 +26,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
```
DanielArndt, line 29: I think this whole file can be reverted now -- it's not used anymore.

DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/534/commits/da289de9084fd2ffe309878323a2346afd3c537d

gruyaume (approved):

PR #535, opened 2024-11-21
feat: Uses JujuFacade in vault_autounseal

saltiyazan:
# Description

- Improves JujuFacade to allow getting a relation by name and to handle secrets without providing a label
- Uses JujuFacade in vault_autounseal.py

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/juju_facade.py
+++ b/lib/charms/vault_k8s/v0/juju_facade.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from typing import List, Literal, cast
 
 from ops.charm import CharmBase
-from ops.framework import Object
 from ops.model import (
     Application,
     ModelError,
@@ -27,7 +26,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 logger = logging.getLogger(__name__)
 
@@ -72,15 +71,22 @@ class SecretAccessDeniedError(FacadeError):
     """Exception raised if the charm does not have permission to access the secret."""
 
 
-class JujuFacade(Object):
+class NoRemoteAppError(FacadeError):
+    """Exception raised when a relation has no remote application."""
+
+
+class MultipleRelationsFoundError(FacadeError):
+    """Exception raised when multiple relations are found."""
+
+
+class JujuFacade:
     """Juju API wrapper class."""
 
     def __init__(self, charm: CharmBase):
-        super().__init__(charm, "juju_facade")
         self.charm = charm
 
     # Secret related methods
-    def get_secret(self, label: str, id: str | None = None) -> Secret:
+    def get_secret(self, label: str | None = None, id: str | None = None) -> Secret:
         """Retrieve a secret and handles error.
 
         Raises:
@@ -89,7 +95,7 @@ class JujuFacade(Object):
         """
         try:
             return (
-                self.charm.model.get_secret(label=label, id=id)
+                self.charm.model.get_secret(id=id)
                 if id
                 else self.charm.model.get_secret(label=label)
             )
@@ -100,7 +106,7 @@ class JujuFacade(Object):
             logger.error("Error getting secret %s: %s", label, e)
             raise TransientJujuError(e) from e
 
-    def secret_exists(self, label: str, id: str | None = None) -> bool:
+    def secret_exists(self, label: str | None = None, id: str | None = None) -> bool:
         """Check if the secret exists.
 
         Raises:
@@ -115,7 +121,7 @@ class JujuFacade(Object):
             raise
 
     def secret_exists_with_fields(
-        self, fields: tuple[str, ...], label: str, id: str | None = None
+        self, fields: tuple[str, ...], label: str | None = None, id: str | None = None
     ) -> bool:
         """Check if the secret exists and has the required fields.
 
@@ -139,7 +145,7 @@ class JujuFacade(Object):
             raise
 
     def _get_secret_content(
-        self, label: str, id: str | None = None, refresh: bool = False
+        self, label: str | None = None, id: str | None = None, refresh: bool = False
     ) -> dict[str, str]:
         try:
             secret = self.get_secret(label=label, id=id)
@@ -153,7 +159,9 @@ class JujuFacade(Object):
             logger.error("Error getting secret content for %s: %s", label, e)
             raise SecretAccessDeniedError(e) from e
 
-    def get_current_secret_content(self, label: str, id: str | None = None) -> dict[str, str]:
+    def get_current_secret_content(
+        self, label: str | None = None, id: str | None = None
+    ) -> dict[str, str]:
         """Get secret content if the secret exists and return currently tracked revision.
 
         Raises:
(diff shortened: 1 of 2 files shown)


lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -207,58 +217,70 @@ def get_secret_content_values(
...
         try:
             secret = self.charm.unit.add_secret(content, label=label)
-            logger.info("Secret %s added to unit", label)
+            logger.info("Secret %s added to unit", label or secret.id)
             return secret
         except ValueError as e:
             logger.error("Invalid secret content %s: %s", content, e)
             raise SecretValidationError(e) from e
 
+    def _create_new_secret(
```
gruyaume, line 240:
```suggestion
    def _create_secret(
```

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -315,6 +337,30 @@ def set_secret_expiry(self, expiry: datetime, label: str, id: str | None = None)
...
+            SecretRemovedError
+        """
+        if secret:
+            secret.grant(relation)
+            return
+        try:
+            secret = self.get_secret(id=secret_id, label=secret_label)
+        except (NoSuchSecretError, SecretRemovedError):
+            raise
+        secret.grant(relation)
```
gruyaume, line 362: The secret is not granted to the relation, but to the app and units on the other side of the relation. The The relation is used to scope the life of the secret.

lib/charms/vault_k8s/v0/vault_autounseal.py
```diff
@@ -363,62 +392,31 @@ def get_outstanding_requests(self, relation_id: int | None = None) -> List[Relat
...
+            if not (
+                credentials_secret_id := self.juju_facade.get_app_relation_data(
+                    self.relation_name, relation_id
+                ).get("credentials_secret_id")
+            ):
+                return False
+            role_id, secret_id = self.juju_facade.get_secret_content_values(
+                "role-id",
+                "secret-id",
+                label=f"{AUTOUNSEAL_CREDENTIALS_SECRET_LABEL_PREFIX}{relation_id}",
```
gruyaume, line 414: Is this a breaking change as we use the secret label now, which didn't exist for prior versions of this charm

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume, line 1: I must say that, looking at the code changes here, I'm not convinced about the introduction of Juju facade. We added an abstraction layer, therefore increasing this charm's complexity, only to reduce the amount of code by 3% (531->515 LOC). Am I missing something?

gruyaume (changes requested):

lib/charms/vault_k8s/v0/vault_autounseal.py
saltiyazan: I agree that this doesn't have a visible code reduction effect here. I think it will be clearer when refactoring the charm code. However the benefits here should be more than just code reduction.
I'll quote the rationale of the spec being implemented. (TE109)
```
The rationale behind this spec is to create standardized, reliable libraries that streamline interactions with Vault and the Juju ecosystem as a whole, ensuring consistent interaction with Juju state that comes from secrets, configurations, and relations. By enforcing clear naming conventions, robust error handling, and idempotency, these libraries aim to simplify charm development and reduce potential for errors.

This change will help with the separation of concerns between writing the configuration logic in the charm, simplifying and stabilizing the interactions with vault, and introducing idempotency and error handling to juju calls.

```

lib/charms/vault_k8s/v0/vault_autounseal.py
saltiyazan: It shouldn't be, we don't rely on this label at all. I just thought it's nicer if the user sees a name associated with the label when running `juju secrets --format=yaml`
knowing that do you still see the risk of this being a breaking change?

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: Renamed.

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: ✅

lib/charms/vault_k8s/v0/vault_autounseal.py
gruyaume: Well yes, if someone deployed Vault prior to this change, this secret wouldn't have a label. Here we are getting the secret based on the label and this method would return False, even though the secret was already created.

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -315,6 +337,30 @@ def set_secret_expiry(self, expiry: datetime, label: str, id: str | None = None)
...
+            NoSuchSecretError
+            SecretRemovedError
+        """
+        if secret:
+            secret.grant(relation)
+            return
+        try:
+            secret = self.get_secret(id=secret_id, label=secret_label)
+        except (NoSuchSecretError, SecretRemovedError):
+            raise
```
gruyaume, line 361: What's the value of such a try/except block?

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -331,97 +377,157 @@ def get_relation_by_id(self, relation_name: str, relation_id: int) -> Relation:
             raise NoSuchRelationError(f"Relation {relation_name}:{relation_id} not found")
         return relation
 
-    def get_relations(self, relation_name: str) -> List[Relation]:
+    def get_relation_by_name(self, relation_name: str) -> Relation:
```
gruyaume, line 380:
```suggestion
    def get_relation_by_name(self, name: str) -> Relation:
```

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -331,97 +377,157 @@ def get_relation_by_id(self, relation_name: str, relation_id: int) -> Relation:
...
-        if not relation:
-            raise NoSuchRelationError(f"Relation {relation_name}:{relation_id} not found")
+        try:
+            relation = (
+                self.get_relation_by_id(relation_name, relation_id)
+                if relation_id
+                else self.get_relation_by_name(relation_name)
+            )
+        except (NoSuchRelationError, MultipleRelationsFoundError):
+            raise
```
gruyaume, line 434: Similar comment, what does this try/except block do?

gruyaume: We have those block all over the place, but we simply re-raise the same error, why catch it at all then?

gruyaume (changes requested):

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: Removed

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: ✅

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: Removed them all. Added the exception to the docstring in public functions where it is needed to signal to the caller the possible exceptions.
More code reduction btw :)

lib/charms/vault_k8s/v0/vault_autounseal.py
saltiyazan: Oh, I didn't mean to have it in the `get` function too. Good catch.
I left however in the `set` just to have a name for the secret.

gruyaume (approved):

PR #539, opened 2024-11-28
chore: Refactor Vault KV and use JujuFacade

saltiyazan:
# Description

This PR has a fix for a potential bug and refactors Vault KV.

- Use JujuFacade in `vault_kv` and its test files.
- Breaking changes in the Vault KV Provider:
- - There is no need for separate setters for relation data. Vault KV details are only valuable together and can be set together.
- - `get_outstanding_kv_requests` is not needed because the lib can't really validate whether a request is fulfilled. Checking if credentials are provided is not enough since a change in the egress subnet of the requirer unit will require the issuance of new credentials to allow the new subnet. (This would have been a potential bug where the provider doesn't issue new credentials to a unit that has its egress subnet changed if the provider misses the `NewVaultKvClientAttachedEvent` event.
- - Removes tests related to `get_outstanding_kv_requests`
- Adds necessary unit tests

# Checklist:


lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -372,8 +373,18 @@ def get_relations(self, relation_name: str, relation_id: int | None = None) -> L
             if not relation_id
             else [self.get_relation_by_id(relation_name, relation_id)]
         )
+        if not relations:
+            logger.warning("No relations found for %s", relation_name)
         return relations
 
+    def get_active_relation_by_id(self, relation_name: str, relation_id: int) -> Relation | None:
```
gruyaume, line 380: The name contradicts with the input, the name suggests we get it by id but we get it by name and id, i'd suggest replacing with `get_active_relation`.

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -502,11 +591,16 @@ def set_app_relation_data(
...
+            relation=relation,
             entity=self.charm.model.app,
         )
 
     def set_unit_relation_data(
-        self, data: dict[str, str], relation_name: str, relation_id: int | None
+        self,
+        data: dict[str, str],
+        relation_name: str,
+        relation_id: int | None = None,
```
gruyaume, line 602: The `relation` prefixes are unnecessary, it's obvious we are talking about the relation. 

The same may apply elsewhere

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -543,6 +638,46 @@ def get_storage_location(self, storage_name: str) -> Path:
...
+
+    def get_egress_subnets(
+        self,
+        relation_name: str,
+        relation_id: int | None = None,
+        relation: Relation | None = None,
+    ) -> List[str]:
+        """Get the list of egress subnets for the given relation.
+
+        This returns how units on the relation will see the charm connecting from.
```
gruyaume, line 669: An example of what this subnet look like would be useful, especially since it's a string.

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -436,57 +476,106 @@ def get_remote_app_relation_data(
...
             )
             raise InvalidRelationDataError(e) from e
 
     def set_app_relation_data(
-        self, data: dict[str, str], relation_name: str, relation_id: int | None
+        self,
+        data: dict[str, str],
+        relation_name: str,
+        relation_id: int | None = None,
+        relation: Relation | None = None,
```
gruyaume, line 578: The `relation` prefixes are unnecessary, it's obvious we are talking about the relation.

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -339,6 +344,7 @@ def __init__(
         relation_name: str,
     ) -> None:
         super().__init__(charm, relation_name)
+        self.juju_facade = JujuFacade(charm)
         self.charm = charm
```
gruyaume, line 348: We probably don't need to set `self.charm`, now that we use juju facade right?

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -447,47 +400,40 @@ def remove_unit_credentials(self, relation: ops.Relation, nonce: str | Iterable[
...
-            ):
-                outstanding_requests.append(request)
-        return outstanding_requests
+        """Get the unit credentials from the relation and load it as a dict."""
+        return json.loads(
+            self.juju_facade.get_app_relation_data(
+                relation_name=self.relation_name,
+                relation=relation,
+            ).get("credentials", "{}")
+        )
```
gruyaume, line 416: The docstring contradicts the method content, is it the unit or the app relation data?

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -497,25 +443,40 @@ def get_kv_requests(self, relation_id: int | None = None) -> List[KVRequest]:
...
+                relation=relation,
+                data={
+                    "mount": mount,
+                    "ca_certificate": ca_certificate,
+                    "vault_url": vault_url,
+                    "credentials": json.dumps(credentials, sort_keys=True),
+                },
+            )
+        except NotLeaderError:
+            return
```
gruyaume, line 470: In some places in this file we have

```python
if not self.juju_facade_is_leader:
    return
do_x()
```

And in other places, like here, we have:

```python
try:
    self.do_x()
except NotLeaderError:
    return
```

Let's stick with one of the approaches

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -447,47 +400,40 @@ def remove_unit_credentials(self, relation: ops.Relation, nonce: str | Iterable[
...
             credentials.pop(n, None)
 
-        relation.data[self.charm.app]["credentials"] = json.dumps(credentials, sort_keys=True)
+        self.juju_facade.set_app_relation_data(
+            relation_name=self.relation_name,
+            relation=relation,
+            data={"credentials": json.dumps(credentials, sort_keys=True)},
+        )
 
     def get_credentials(self, relation: ops.Relation) -> dict:
```
gruyaume, line 409: This may not be quite relevant to this PR but it feels like get_credentials should return a dict, the structure of this dict will always be the same

lib/charms/vault_k8s/v0/vault_kv.py
```diff
@@ -497,25 +443,40 @@ def get_kv_requests(self, relation_id: int | None = None) -> List[KVRequest]:
...
-        if not relation:
-            return False
+    def set_kv_data(
+        self,
+        relation: ops.Relation,
+        mount: str,
+        ca_certificate: str,
+        vault_url: str,
+        nonce: str,
+        secret_id: str,
```
gruyaume, line 453: Is this the juju secret id or the vault secret id? Those two can easily be confused, I'd recomment being explicit here

src/charm.py
```diff
@@ -973,27 +986,27 @@ def _get_backup_key(self) -> str:
...
-            egress_subnets: egress subnet
-        """
-        self.vault_kv.set_mount(relation, mount)
-        vault_url = self._get_relation_api_address(relation)
-        self.vault_kv.set_ca_certificate(relation, ca_certificate)
-        self.vault_kv.set_egress_subnets(relation, egress_subnets)
-        if vault_url is not None:
-            self.vault_kv.set_vault_url(relation, vault_url)
+        role_name: str,
+        secret_id: str,
```
gruyaume, line 995: Similar comment, I'd suggest being quite explicit as to whether this is the vault secret id or the juju secret id

gruyaume (changes requested):

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: Fixed

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: Removed

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: app, added.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: Good point, there is no need to have both in this file.

lib/charms/vault_k8s/v0/vault_kv.py
saltiyazan: I'm not sure I get this comment, the function returns a dict.

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -330,7 +331,7 @@ def grant_secret(
         secret.grant(relation)
 
     # Relation related methods
-    def get_relation_by_id(self, relation_name: str, relation_id: int) -> Relation:
+    def get_relation_by_id(self, name: str, id: int) -> Relation:
```
gruyaume, line 334: The method name does not match its content, if we get it by id, why ask for the name? I'd change to `get_relation`, just like we have `get_relations` below (and not `get_relations_by_id`)

lib/charms/vault_k8s/v0/juju_facade.py
```diff
@@ -422,71 +452,108 @@ def get_remote_app_relation_data(
...
         return relation.data.get(relation.app, {})
 
     def get_unit_relation_data(
-        self, relation_name: str, relation_id: int | None
+        self,
+        name: str,
+        id: int | None = None,
+        relation: Relation | None = None,
     ) -> RelationDataContent | dict[str, str]:
         """Get relation data from the remote unit databag.
```
gruyaume, line 472: The docstring does not seem quite right, this would not be the remote unit

lib/charms/vault_k8s/v0/vault_kv.py
gruyaume: Oops, I meant an object (instead of a dict) - edited

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: ✅

lib/charms/vault_k8s/v0/juju_facade.py
saltiyazan: ✅

gruyaume (approved):

PR #542, opened 2024-11-28
build: Use lxd 6/stable

DanielArndt:
Upgrade from lxd 5.0 to lxd 6.x.

5.0 doesn't have a channel that follows the latest minor version, but 6 does, so using 6/stable will get us the latest updates, _and_ won't break with every update.

# Checklist:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup LXD
         uses: canonical/setup-lxd@main
         with:
-          channel: 5.20/stable
+          channel: 6/stable
 
       - name: Install charmcraft
         run: sudo snap install charmcraft --classic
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Setup LXD
         uses: canonical/setup-lxd@main
         with:
-          channel: 5.0/stable
+          channel: 6/stable
```
ghislainbourgeois, line 15: I don't think we should use `6/stable`. In my other PR, I put `5.21/stable`, and that is the LTS release of LXD. The versioning scheme of LXD is to take the latest minor release of a major version as the LTS. With `6/stable`, there will be a lot of new features and potential breaking changes.

DanielArndt: I thought we were moving from 5.xx/stable to 5.0/stable so that the builds didn't break every time a new version?

@saltiyazan was asking about this and I may have misinformed him.

DanielArndt: OK, I did some more reading and I see what you're saying. It seems they locked in 5.21 as an LTS, although the last LTS was 5.0. So, we were probably reverting to 5.0 (and never finished) before 5.21 came out.

PR #559, opened 2024-12-09
feat: Implement PKI Manager

DanielArndt:
This implements the PKI manager and moves associated code there.

It also introduces some new tests, and some minor refactoring in the implementation to improve readability (although, I think it could still use some work).

No functional changes (at least, that's the intention).

# Checklist:

### DIFF
--- a/lib/charms/vault_k8s/v0/vault_managers.py
+++ b/lib/charms/vault_k8s/v0/vault_managers.py
@@ -864,7 +864,7 @@ class PKIManager:
         self._vault_pki = vault_pki
         self._tls_certificates_pki = tls_certificates_pki
 
-    def intermediate_ca_cert_request(self) -> CertificateRequestAttributes:
+    def get_intermediate_ca_cert_request(self) -> CertificateRequestAttributes:
         """Return the certificate request attributes for the intermediate CA.
 
         Returns:
@@ -872,7 +872,7 @@ class PKIManager:
             intermediate CA certificate from the TLS certificates provider.
         """
         return CertificateRequestAttributes(
-            self._common_name,
+            common_name=self._common_name,
             is_ca=True,
         )
 
@@ -889,7 +889,7 @@ class PKIManager:
         This is the CA certificate that the provider charm has issued to Vault.
         """
         provider_certificate, private_key = self._tls_certificates_pki.get_assigned_certificate(
-            certificate_request=self.intermediate_ca_cert_request()
+            certificate_request=self.get_intermediate_ca_cert_request()
         )
         if not provider_certificate:
             logger.debug("No intermediate CA certificate available")
--- a/lib/charms/vault_k8s/v0/vault_managers.py
+++ b/lib/charms/vault_k8s/v0/vault_managers.py
@@ -838,7 +838,7 @@ class PKIManager:
         self,
         charm: CharmBase,
         vault_client: VaultClient,
-        common_name: str,
+        certificate_request_attributes: CertificateRequestAttributes,
         mount_point: str,
         role_name: str,
         vault_pki: TLSCertificatesProvidesV4,
@@ -849,37 +849,21 @@ class PKIManager:
         Args:
             charm: The charm this manager is associated with
             vault_client: The Vault client object
-            common_name: The common name for the certificates allowed by the PKI backend
+            certificate_request_attributes: The certificate request attributes
+                that were used when requesting an intermediate certificate from
+                the tls_certificates_pki relation provider.
             mount_point: The mount point in Vault for the PKI backend
             role_name: The role name for the PKI backend
             vault_pki: The vault_pki provider relation helper library
             tls_certificates_pki: The tls_certificates_pki requirer relation helper library
         """
-        self.validate_input(common_name=common_name)
         self._vault_client = vault_client
         self._juju_facade = JujuFacade(charm)
-        self._common_name = common_name
         self._mount_point = mount_point
         self._role_name = role_name
         self._vault_pki = vault_pki
         self._tls_certificates_pki = tls_certificates_pki
-
-    def get_intermediate_ca_cert_request(self) -> CertificateRequestAttributes:
-        """Return the certificate request attributes for the intermediate CA.
-
-        Returns:
-            The CertificateRequestAttributes that should be used to request the
-            intermediate CA certificate from the TLS certificates provider.
-        """
-        return CertificateRequestAttributes(
-            common_name=self._common_name,
-            is_ca=True,
-        )
-
-    def validate_input(self, *, common_name: str):
-        """Validate the inputs to this class."""
-        if common_name == "":
-            raise ManagerError("Common name cannot be empty.")
+        self._certificate_request_attributes = certificate_request_attributes
 
     def _get_pki_intermediate_ca_from_relation(
         self,
@@ -889,7 +873,7 @@ class PKIManager:
         This is the CA certificate that the provider charm has issued to Vault.
         """
         provider_certificate, private_key = self._tls_certificates_pki.get_assigned_certificate(
-            certificate_request=self.get_intermediate_ca_cert_request()
+            certificate_request=self._certificate_request_attributes
         )
         if not provider_certificate:
             logger.debug("No intermediate CA certificate available")
@@ -970,12 +954,12 @@ class PKIManager:
         if not self._vault_client.is_common_name_allowed_in_pki_role(
             role=self._role_name,
             mount=self._mount_point,
-            common_name=self._common_name,
+            common_name=self._certificate_request_attributes.common_name,
         ) or issued_certificates_validity != self._vault_client.get_role_max_ttl(
             role=self._role_name, mount=self._mount_point
         ):
             self._vault_client.create_or_update_pki_charm_role(
-                allowed_domains=self._common_name,
+                allowed_domains=self._certificate_request_attributes.common_name,
                 mount=self._mount_point,
                 role=self._role_name,
                 max_ttl=f"{issued_certificates_validity}s",
@@ -993,7 +977,6 @@ class PKIManager:
         except (VaultClientError, IndexError) as e:
             logger.error("Failed to get the first issuer: %s", e)
             return
-
         try:
             issuers_config = self._vault_client.read(path=f"{self._mount_point}/config/issuers")
             if issuers_config and not issuers_config["default_follows_latest_issuer"]:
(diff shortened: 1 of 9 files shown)


lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -816,3 +829,243 @@ def get_provider_vault_token(
...
+        """
+        return CertificateRequestAttributes(
+            self._common_name,
+            is_ca=True,
+        )
+
+    def validate_input(self, *, common_name: str):
+        """Validate the inputs to this class."""
+        if common_name == "":
+            raise ManagerError("Common name cannot be empty.")
```
DanielArndt, line 882: I'm indifferent about this, but I wanted it to error if it was empty. It actually saved me once already.

An assert or just a simple check would work, but I thought I would start something extensible in case there's other things we wanted to validate.

src/charm.py
```diff
@@ -331,7 +329,7 @@ def _configure(self, _: EventBase) -> None:  # noqa: C901
             return
         if not (vault := self._get_active_vault_client()):
             return
-        self._configure_pki_secrets_engine()
+        self._configure_pki_secrets_engine(vault)
```
DanielArndt, line 332: We could probably combine `configure()` with `sync()` (I haven't tried it, but I don't see any reason we can't).

I left it as-is to not change the behaviour too much here, but if there is support for combining them let me know.

src/charm.py
```diff
@@ -400,122 +398,22 @@ def _on_vault_kv_client_detached(self, event: VaultKvClientDetachedEvent):
...
-        self,
-    ) -> Tuple[ProviderCertificate | None, PrivateKey | None]:
-        """Get the intermediate CA certificate."""
-        certificate_request = self._get_certificate_request()
-        if not certificate_request:
-            logger.error("Certificate request is not valid")
-            return None, None
-        provider_certificate, private_key = self.tls_certificates_pki.get_assigned_certificate(
-            certificate_request=certificate_request
+        logger.debug("Common name: %s", common_name)
```
DanielArndt, line 406: Left this in. Needs to be removed.

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -816,3 +829,243 @@ def get_provider_vault_token(
...
+        self.validate_input(common_name=common_name)
+        self._vault_client = vault_client
+        self._juju_facade = JujuFacade(charm)
+        self._common_name = common_name
+        self._mount_point = mount_point
+        self._role_name = role_name
+        self._vault_pki = vault_pki
+        self._tls_certificates_pki = tls_certificates_pki
+
+    def intermediate_ca_cert_request(self) -> CertificateRequestAttributes:
```
gruyaume, line 867: It feels like this func is missing a prefix (ex. `get_intermediate_ca_cert_request`)

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -816,3 +829,243 @@ def get_provider_vault_token(
...
+        """Return the certificate request attributes for the intermediate CA.
+
+        Returns:
+            The CertificateRequestAttributes that should be used to request the
+            intermediate CA certificate from the TLS certificates provider.
+        """
+        return CertificateRequestAttributes(
+            self._common_name,
```
gruyaume, line 875: A bit of a nit but I wouldn't necessarily assume that the common name is the first parameter (even though it is), I'd encourage using `common_name=self._common_name`

lib/charms/vault_k8s/v0/vault_managers.py
gruyaume, line 882: I'm not sure that it should be up to vault managers to validate the common name, we should likely leave that up to the tls certificates library.

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -816,3 +829,243 @@ def get_provider_vault_token(
...
+            common_name: The common name for the certificates allowed by the PKI backend
+            mount_point: The mount point in Vault for the PKI backend
+            role_name: The role name for the PKI backend
+            vault_pki: The vault_pki provider relation helper library
+            tls_certificates_pki: The tls_certificates_pki requirer relation helper library
+        """
+        self.validate_input(common_name=common_name)
+        self._vault_client = vault_client
+        self._juju_facade = JujuFacade(charm)
+        self._common_name = common_name
```
gruyaume, line 861: Should we have the CertificateRequestAttributes be the parameter instead of the common_name? This will allow requesting certs with more fields than just the common name without changing anything here.

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -816,3 +829,243 @@ def get_provider_vault_token(
...
+            return False
+        certificate_validity = (
+            intermediate_ca_certificate.expiry_time
+            - intermediate_ca_certificate.validity_start_time
+        )
+        certificate_validity_seconds = certificate_validity.total_seconds()
+        return certificate_validity_seconds > current_ttl
+
+    def configure(self):
+        """Perform the initial configuration of the PKI backend."""
```
gruyaume, line 935: Not just initial right? And I wouldn't make assumptions as to when this method is called (as it's outside of this function's control), I would simply encourage to state what it does.

lib/charms/vault_k8s/v0/vault_managers.py
```diff
@@ -816,3 +829,243 @@ def get_provider_vault_token(
...
+            if issuers_config and not issuers_config["default_follows_latest_issuer"]:
+                logger.debug("Updating issuers config")
+                self._vault_client.write(
+                    path=f"{self._mount_point}/config/issuers",
+                    data={
+                        "default_follows_latest_issuer": True,
+                        "default": first_issuer,
+                    },
+                )
+        except (TypeError, KeyError):
```
gruyaume, line 1008: Are we expecting those errors in the `read` or in the `write`. I would encourage making the try/except as small as possible so that the log messages in the except matches with the specific try - This helps in debugging.

src/charm.py
```diff
@@ -400,122 +398,21 @@ def _on_vault_kv_client_detached(self, event: VaultKvClientDetachedEvent):
...
+            tls_certificates_pki=self.tls_certificates_pki,
         )
-        if not provider_certificate:
-            logger.debug("No intermediate CA certificate available")
-            return None, None
-        if not private_key:
-            logger.debug("No private key available")
-            return None, None
-        return provider_certificate, private_key
+        manager.configure()
```
gruyaume, line 415: I'm a bit confused by the fact that there is both a `configure` and a `sync` methods in this manager. Is this really necessary?

gruyaume (changes requested):

src/charm.py
DanielArndt: see https://github.com/canonical/vault-k8s-operator/pull/559#discussion_r1890166870

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: this code was moved from elsewhere. I can try to narrow it down if we feel that's worthwhile, but I was trying to avoid functional changes as much as possible.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: see https://github.com/canonical/vault-k8s-operator/pull/559#discussion_r1889101177

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: That probably makes sense. let me take a peek.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Yeah this was a bad one. I put it in as a placeholder, because there was nothing in the original function it was moved from, and then never revisited it.

I'll write something a little better here.

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/559/commits/da96e942559dc99191f384f51ce356d7a7c510dc

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/559/commits/da96e942559dc99191f384f51ce356d7a7c510dc

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: This becomes irrelevant if we do https://github.com/canonical/vault-k8s-operator/pull/559#discussion_r1890760925

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Definitely makes more sense, since the manager it isn't the thing putting the request in the databag in the first place. I'll make this change.

DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/559/commits/04162cb3c26d73c578a49092153cba2a3e30c5fd

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: Irrelevant now because of https://github.com/canonical/vault-k8s-operator/pull/559/commits/04162cb3c26d73c578a49092153cba2a3e30c5fd

lib/charms/vault_k8s/v0/vault_managers.py
DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/559/commits/75059d5fc72815ba4e8ea54a0eba1c40bdb5704f

tests/unit/lib/charms/vault_k8s/v0/test_vault_managers.py
```diff
@@ -1,19 +1,40 @@
...
+
+SECONDS_IN_HOUR = 3600
+
+
+def is_error_logged(caplog: pytest.LogCaptureFixture, error_message: str):
+    error_records = [record for record in caplog.records if record.levelname == "ERROR"]
+    return any(error_message in record.message for record in error_records)
 
 
-class TestAutounsealRequirerManager:
```
DanielArndt, line 16: Oof, bad merge here. Need to fix.

DanielArndt: https://github.com/canonical/vault-k8s-operator/pull/559/commits/b2df951e7e1ade47c97cf55a5a69daa8c196948e

gruyaume (approved):

PR #577, opened 2025-01-07
feat: Implements the backup manager

DanielArndt:
# Checklist:

- [x] I have bumped the version of any required library.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,7 +10,6 @@ For more information on Vault, please visit https://www.vaultproject.io/.
 import json
 import logging
 import socket
-from datetime import datetime
 from typing import Any, Dict, List
 
 import hcl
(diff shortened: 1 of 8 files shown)


src/charm.py
```diff
@@ -844,42 +707,31 @@ def _set_pebble_plan(self) -> None:
...
-
-        Args:
-            snapshot: Snapshot to be restored as a StreamingBody from the S3 storage.
+    def _get_active_vault_client(self) -> VaultClient | None:
+        """Return a client for the _active_ vault service.
 
-        Returns:
-            bool: True if the restore was successful, False otherwise.
+        This may not be the Vault service running on this unit.
         """
```
DanielArndt, line 714: I wonder if there is a way to synchronize the juju leader with the vault leader.

If there was, we wouldn't need this logic since we could only allow this action to run on the leader (and hence, avoid the transmission of the snapshot to another unit). Or, should we force them to run this on the unit that is the Vault leader?

ghislainbourgeois: I do not think there is a way to synchronize the juju leader with the vault leader, no.

I believe this is the simplest approach for users. I would keep it that way, even if it is less efficient.

ghislainbourgeois (approved):

PR #578, opened 2025-01-08
feat: Implements Vault Helpers

saltiyazan:
# Description

This PR moves functions related to the config of the Vault service that are used by both charms (k8s and machine) from the charm code to the `vault_helpers` library that will be shared by the charms.

# Checklist:

- [x] I have bumped the version of any required library.


lib/charms/vault_k8s/v0/vault_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            if vault.is_sealed():
+                logger.error("Can't remove node from cluster - Vault is sealed")
+                return
+        except VaultClientError as e:
+            logger.error("Can't remove node from cluster - Vault status check failed: %s", e)
+            return
+        return vault
+
+    def get_active_vault_client(self, api_address: str, ca_cert_path: str) -> VaultClient | None:
```
DanielArndt, line 87: Not new in this PR, but I got bit by the naming of this, because it is not necessarily the client for the _active_  vault.

Instead, it is the vault client for the current unit, if it is active or standby (i.e the service is fully initiallized).

In restore backup, we need the client for the  _active_ vault, which you would expect this function to provide you.

If we move forward, we should consider some better naming for these.

DanielArndt (commented):
I'm not against this, but it also doesn't give me a warm fuzzy feeling. Originally, I had pictured this mostly being things like `get_approle_auth_secret` which wraps juju_facade with some more Vault-charm specific logic, but I guess there's not much of that.

I just don't know if a charm library is overkill specifically for sharing code. Then again, I suppose it's really not that different from the managers.

I'd be interested to hear what others think.

lib/charms/vault_k8s/v0/vault_helpers.py
saltiyazan: Moved these parts to #581 and reduced this PR to contain config file related functions only as we discussed.

src/charm.py
```diff
@@ -630,7 +628,9 @@ def _generate_vault_config_file(self) -> None:
 
         autounseal_configuration_details = self._get_vault_autounseal_configuration()
 
-        content = _render_vault_config_file(
+        content = render_vault_config_file(
+            config_template_path=CONFIG_TEMPLATE_DIR_PATH,
+            config_template_name=CONFIG_TEMPLATE_NAME,
```
DanielArndt, line 633: We could inline the templates directly in the `vault_helpers` as we did for the managers.

There's benefits to both. If we put them inline, we don't need to repeat them multiple places. If we keep them as seperate files, you tend to get better IDE support like syntax highlighting.

Just a thought, your call.

DanielArndt (approved):

PR #583, opened 2025-01-10
docs: Expand CONTRIBUTING docs

DanielArndt:
Add details on

- Library structure
- Running integration tests locally

# Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,27 @@ uv sync
 source .venv/bin/activate
 ```
 
+## Library Structure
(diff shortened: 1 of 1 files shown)


DanielArndt:
@saltiyazan Can you take a peak and see if this is correct?

DanielArndt:
> It looks good, however I probably would avoid listing all libraries while making sure it is clear that some of the libraries owned by this charm are only for internal use of the Vault charms.

I agree. I struggled with this a bit. I need to find a better approach, I think.

PR #606, opened 2025-03-06
feat: Implements the requirer side of ingressPerUnit

saltiyazan:
# Description

Implements the requirer side of ingressPerUnit
When Vault is only accessible behind an ingress the user needs to access each unit individually to unseal for example.

# Checklist:

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -85,7 +85,14 @@ requires:
     interface: vault-autounseal
     limit: 1
   ingress:
+    description: |
+      Ingress relation for the vault app (ingress per app).
     interface: ingress
+    limit: 1
+  ingress-per-unit:
+    description: |
+      Ingress relation for the vault units (ingress per unit).
+    interface: ingress_per_unit
     limit: 1
   tls-certificates-access:
     interface: tls-certificates
--- a/lib/charms/traefik_k8s/v1/ingress_per_unit.py
+++ b/lib/charms/traefik_k8s/v1/ingress_per_unit.py
@@ -1,0 +1,884 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+r"""# Interface Library for ingress_per_unit.
+
+This library wraps relation endpoints using the `ingress_per_unit` interface
+and provides a Python API for both requesting and providing per-unit
+ingress.
+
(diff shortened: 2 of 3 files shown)


charmcraft.yaml
```diff
@@ -85,8 +85,15 @@ requires:
     interface: vault-autounseal
     limit: 1
   ingress:
+    description: |
+      Ingress relation for the vault app (ingress per app).
```
gruyaume, line 89: Why do we support both ingress-per-app and ingress-per-unit. Is this standard in the ecosystem? Or should we more opinionated and tell people which to use.

gruyaume: > if they need ingress and they should be instructed to use ingress-per-unit too (or alone)
> if their deployment requires accessing each unit separately behind the ingress.

I'm concerned about those `ifs`. Let's be prescriptive around the use of ingress, especially in our [production blueprint](https://charmhub.io/vault-k8s/docs/h-production-blueprint). We know that admin users will need to access each unit separately, and we know that it would be more reliable and simpler for regular users to access Vault through a standalone endpoint. We tell users to deploy Vault with 5 units, alongside the COS stack and with s3 storage. My question is, "What is our recommendation around ingress?". 

From the last message, it's more apparent that we need both integrations. But it's still unclear whether we would recommend deploying one Traefik in the same model as Vault with two integrations (`ingress-per-unit` and `ingress-per-app`) or two Traefiks with one relation to Vault each.

saltiyazan: Recommendation:
- Deploy 5 units of Vault
- Deploy Traefik (We will use the same Traefik for both integrations)
- Integrate Vault to Traefik with the `ingress` (`ingress-per-app`) integration to access Vault through a single endpoint.
- Integrate Vault to Traefik with the `ingress-per-unit` integration to access each unit individually through the ingress (Optional)

This is what I have in mind. Are there any details I am missing?

gruyaume (approved):

gruyaume:
@saltiyazan can we go ahead and merge this?

saltiyazan:
> @saltiyazan can we go ahead and merge this?

Yes I want to merge it, I just keep getting other PRs merged before so I have to update the branch.
I enabled auto-merge. Will update the docs.

PR #612, opened 2025-03-19
test: Pin self-signed certificates revision

DanielArndt:
Add assertion for CA certificate presence and pin self-signed certificates revision

Also, remove an unnecessary wait.

# Checklist:

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -34,6 +34,7 @@ APPLICATION_NAME = "vault-k8s"
 LOKI_APPLICATION_NAME = "loki-k8s"
 PROMETHEUS_APPLICATION_NAME = "prometheus-k8s"
 SELF_SIGNED_CERTIFICATES_APPLICATION_NAME = "self-signed-certificates"
+SELF_SIGNED_CERTIFICATES_REVISION = 263
 VAULT_KV_REQUIRER_1_APPLICATION_NAME = "vault-kv-requirer-a"
 VAULT_KV_REQUIRER_2_APPLICATION_NAME = "vault-kv-requirer-b"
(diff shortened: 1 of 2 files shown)


tests/integration/test_integration.py
```diff
@@ -34,6 +34,7 @@
 LOKI_APPLICATION_NAME = "loki-k8s"
 PROMETHEUS_APPLICATION_NAME = "prometheus-k8s"
 SELF_SIGNED_CERTIFICATES_APPLICATION_NAME = "self-signed-certificates"
+SELF_SIGNED_CERTIFICATES_REVISION = 263
```
gruyaume, line 37: Why pin to a specific revision and not use the actual revision that people will used when deploying Vault (i.e. whatever the latest revision on `1/stable`?

gruyaume (changes requested):

tests/integration/test_integration.py
DanielArndt: To make the point more real -- this was a very real problem with self-signed-certificates and tls-certificates-requirer. It took quite a while to figure out which revisions were compatible with which as I detangled the V4 updates. This required me to write a totally separate version of the integration tests that I was using to test my backports, because the integration tests for v3 of the lib would no longer work with the latest version of the tls-certificates requirer, and vice-versa. This would not have been an issue if the revisions were pinned.

gruyaume (approved):

tests/integration/test_integration.py
DanielArndt: I agree we could be more consistent. I'll add a story to the backlog to follow up on this.

DanielArndt: https://warthogs.atlassian.net/browse/TLSENG-681

PR #615, opened 2025-03-21
chore: reorganize vault charms into monorepo

gruyaume:
# Description

This repository now contains both the K8s and Machine charm source code. Here is the new file structure:
- `k8s`: Kubernetes Charm
- `machine`: Machine Charm 

## Sharing common code between the K8s and Machine charms

The Vault machine and K8s charms share a lot of code. The K8s charm owns this common code and you can fetch this code into the Machine charm using:

```shell
make vendor-libs
```

This command will copy the code from the k8s charm's `lib/vault/` directory to the machine charm's `lib/vault/` directory.

**We do not use charm libraries to share code between the 2 Vault charms anymore**

## Removing Vault KV library dependency on Juju facade

Of all the libraries this charm owns, only Vault KV is used by other charms. We recently added a dependency on the Juju facade for this library (#539), and I believe adding this dependency was a bad idea, and I'm removing this dependency here. The rationale is that:
1. There is no clean way to represent those charm library dependencies. Our KV users would have had to fetch the Vault KV lib and try to use it, then see an error and understand that this charm lib depends on another one.
2. It exposed our internal implementation details and unnecessarily complexified the life of charm developers trying to integrate to Vault KV

This update to Vault KV (and the fact that Juju facade is not a charm library anymore) is the only notable change part of this PR outside of the restructuring. The reason why I included it part of this PR is that I moved all common code between Vault charms into the away from `lib/charms/vault_k8s` and into the `lib/vault` directory. This makes Juju facade, a lib only used in Vault charms, unavailable to outside users.

## Workflow changes

Our workflows now run tests for the K8s and machine charms.

## I realize this is an impossibly huge change

I realize this is an impossibly huge change. I would encourage reviewers to clone the code and try it for yourself. This change does not include any change to the Vault source code or its tests.

## Rationale

While trying to create a new release of the Vault charm (1.17), I realized it wasn't possible to create new major releases of Vault charm libraries without painting ourselves in a corner. We would have had the `1.16` branch of Vault K8s own the `v0` of the lib and the `main` branch owns the `v1` of the library. After discussing with @DanielArndt and @ghislainbourgeois , we decided now was the best time to merge the Vault repos into 1.

# Checklist:


machine/charmcraft.yaml
```diff
@@ -0,0 +1,173 @@
...
+  network encryption-as-a-service, or generate AWS IAM/STS
+  credentials, SQL/NoSQL databases, X.509 certificates,
+  SSH credentials, and more.
+links:
+  website:
+    - https://charmhub.io/vault
+  source:
+    - https://github.com/canonical/vault-operator
+  issues:
+    - https://github.com/canonical/vault-operator/issues
```
ghislainbourgeois, line 22: This will probably need to change also, but maybe we can wait until this is merged and we rename the repository. WDYT?

ghislainbourgeois (approved):
I left 2 small questions, but LGTM

machine/charmcraft.yaml
gruyaume: Good catch, done here.

.github/workflows/main.yaml
```diff
@@ -11,43 +11,116 @@ concurrency:
...
+  k8s-publish-charm:
+    name: Publish K8s Charm
+    needs:
+      - k8s-lint-report
+      - k8s-grafana-dashboard-lint-report
+      - k8s-static-analysis
+      - k8s-unit-tests-with-coverage
+      - k8s-integration-test
+    if: ${{ github.ref_name == 'main' || startsWith(github.ref_name, 'release-') }}
+    uses: canonical/identity-credentials-workflows/.github/workflows/publish-charm.yaml@dev-path
```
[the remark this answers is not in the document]
gruyaume: Yes, good catch. I changed it in the latest commit.

PR #665, opened 2025-07-04
feat: Expose the sign-verbatim endpoint

saltiyazan:
# Description

**Note**: This PR is work in progress as it lacks tests. For now the purpose is to get feedback from the team whether to continue or not.

This PR exposes the [sign-verbatim](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-verbatim) endpoint of Vault. The users will be able to configure Vault PKI to allow sign-verbatim through a charm config option. When set to true no other PKI configs are considered or needed. This will make Vault PKI sign the CSR as is with all its attributes.
Vault discourages using this endpoint it says: "This is a potentially dangerous endpoint and only highly trusted users should have access."

### Motivation:
Some users want to have a certificate that matches their csr when using Vault as a cert provider. By default Vault will not respect the subject name attributes from the csr (organization, organizational unit, country...) and will use those as configured in the PKI role.
Sign verbatim will generate a certificate that includes those fields just like they were requested in the CSR.

### Risk:
- Requirers can abuse this and ask for values that shouldn't be allowed, for example the requirer asks for an org that doesn't match their actual org, Vault can't validate that.
- Reverting this change in the future would be a breaking change for the API

### Opinion:
The default value of this config is false and we have a clear warning copied from the Vault docs in our charmcraft.yaml. If the operator decides to set this to true despite the warning they probably know what they are doing and it won't be any less secure than a solution like `self-signed-certificates` that does exactly that.

# Checklist:

### DIFF
--- a/k8s/charmcraft.yaml
+++ b/k8s/charmcraft.yaml
@@ -389,6 +389,12 @@ config:
       description: >-
         This value specifies the L (Locality) value in the Subject field of the certificate issued by Vault PKI.
         If not set the issued certificate will not have an L value in the Subject field.
+    pki_sign_verbatim:
+      type: boolean
+      default: false
+      description: >-
+        If set to true, all other pki configuration options will be ignored and the Vault charm will sign the certificate verbatim.
+        Using this option is not recommended and should only be done when the requirers are trusted.
 
 # ACME engine options
     acme_ca_common_name:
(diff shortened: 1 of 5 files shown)


saltiyazan:
Please share your thoughts on whether this is useful to have.
@DanielArndt, @kayra1, @ghislainbourgeois, @gruyaume

PR #701, opened 2025-07-22
docs: Update production blueprint

saltiyazan:
# Description
Update production blueprint to recommend the usage of ingress and separate machine from k8s

# Checklist:


docs/images/production_blueprint_k8s.png
gruyaume, line 1: I don't think we want to use the Traefik from the observability model, it does not make much sense. I'd rather have a Traefik in the same model as Vault.

docs/images/production_blueprint_machine.png
gruyaume, line 1: Why a separate model for ingress? We probably want Traefik in the same model as Vault.

gruyaume, line 1: The size of the white box for cos lite is too big for its content (green box).

gruyaume (changes requested):

docs/images/production_blueprint_k8s.png
saltiyazan: Good point, but Traefik comes with the bundle, so I removed the label `cos-lite`, now I refer to this as cos tools

gruyaume: I don't quite understand the point, there can be multiple ingresses out there. I would keep the `cos-lite` name, `cos tools` is not a thing.

docs/images/production_blueprint_machine.png
gruyaume: Yes an org *could* have a different model for ingress, but here we are presenting how we recommend users deploy our solution. Just like with `cos-lite` it is quite reasonable to have the ingress in the same model, I don't see what we gain by asking users to create a separate model just for that. Just like with grafana-agent, in reality, it could have been in a different model but what do users gain? It just complicates the deployment.

docs/images/production_blueprint_k8s.png
gruyaume, line 1: We should use `ingress-per-unit`, not `ingress-per-app`

gruyaume, line 1: Let's use a line or have the arrows represent the provider/requirer meaning

saltiyazan: done

saltiyazan: done

gruyaume (approved):

PR #707, opened 2025-08-08
feat: use systemd to encrypt the external vault token at rest

DanielArndt:
# Description

This change allows us to pass the token for the external vault into the service using [systemd credentials](https://systemd.io/CREDENTIALS/).

Although this token is eventually read into an environment variable, just as before, it is now encrypted on disk. Furthermore, it opens up the possibility to read this token directly from the file-like object in later versions of Vault (indirect values are supported in v1.19+).

# Checklist:


machine/tests/integration/config.py
```diff
@@ -11,7 +11,7 @@
 MINIO_APPLICATION_NAME = "minio"
 MINIO_S3_ACCESS_KEY = "vaultintegrationtest"
 MINIO_S3_SECRET_KEY = "vaultintegrationtest"
-NUM_VAULT_UNITS = 3
+NUM_VAULT_UNITS = 1
```
saltiyazan, line 14: Why are we making this change?

saltiyazan (changes requested):
Nice work, this looks good.
I'm just not sure why we have reduced the number of units we do integration test with.

machine/tests/integration/config.py
DanielArndt: Oops, this was just a change I made to test quicker -- and accidentally committed. Embarassing. :cry:

saltiyazan (approved):

PR #977, opened 2026-05-25
feat: Implement self-signed CA to enable the PKI engine with spec-kit

### LINKED ISSUE #839: Why not allow to store the root secret in Vault itself? "tls-certificates-pki relation is missing, cannot configure PKI secrets engine"
### Bug Description

Somehow, the later gen of Vault charm removed the ability to act as the root CA for TLS consumer charms although it's not called out in the document:
https://github.com/canonical/vault-k8s-operator/blob/main/docs/reference/key_differences_between_charm_versions.md

Previously we were able to use the charm deployed Vault as the secret store and the root CA at the same time. e.g.
- https://ubuntu.com/kubernetes/charmed-k8s/docs/using-vault#deploying-charmed-kubernetes-with-vault-as-a-root-ca
- https://docs.openstack.org/charm-guide/latest/admin/security/tls.html#certificate-management-with-vault

However, if we deploy charmed vault with later gen, it's stuck at:
```
Unit                          Workload  Agent  Machine  Public address  Ports     Message
tls-certificates-requirer/0*  active    idle   1        10.0.9.140                0/1 certificate requests are fulfilled
vault/0*                      blocked   idle   0        10.0.9.191      8200/tcp  tls-certificates-pki relation is missing, cannot configure PKI secrets engine
```

There is a document about how to use charmed vault as an intermediate CA, but there is no document about how to use it as root CA.
https://github.com/canonical/vault-k8s-operator/blob/main/docs/how-to/use_as_intermediate_ca.md

If we follow the doc above, there is a question mark around the security and secrecy. Because, the root CA (with the super power) is generated on the self-signed-certificates unit and saved to Juju secret storage. However, if we don't have Vault in the first place, the super power CA is saved as plain text in Juju database. But if we can enable the root CA in charmed vault itself, then the secret key never leaves the vault unit and the key is saved encrypted in the vault raft storage backend by leveraging the function of Vault itself. So it can act as an egg for another Juju controller to secure Juju secret storage to break a chicken-and-egg situation.

### To Reproduce

1.
```
juju deploy vault --channel 1.18/edge \
    --config pki_ca_common_name=mydomain.example.com --config pki_allow_subdomains=true
````
2.
```
juju deploy tls-certificates-requirer --base ubuntu@22.04 --channel latest/edge \
    --config common_name=demo.mydomain.example.com  --config sans_dns=demo.mydomain.example.com
```

3.
```
juju relate tls-certificates-requirer:certificates vault:vault-pki
```

### Environment

1.18/edge    589

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

saltiyazan:
# Description

Fixes #839 

Allows the activation of the PKI engine and the usage of Vault as a self-signed root CA.

# Checklist:


VercinGor (approved):
Look good. As discussed on the side, I'm still on the fence on what all to include from the specs into the git history. Lets kick this rolling and we will iterate over and see what works best as we go.

PR #1005, opened 2026-06-19
chore: Use the latest shared workflows using just

saltiyazan:
# Description

- Use the latest shared workflows using just
- We build and test on each commit for the most common path only AMD@26.04 this saves 8 self hosted runners per push. The rest of the tests and checks run on merge, or if the PR is labeled with `test-arm` or `test-s390x`

# Checklist:


VercinGor (approved):
looks in order

.github/workflows/grafana-dashboard-lint.yaml
kayra1, line 1: there is a just recipe for this

.github/workflows/main.yaml
```diff
@@ -29,49 +32,64 @@ jobs:
...
       path: k8s
       artifact-suffix: k8s
 
   k8s-build-arm:
-    needs:
-      - k8s-lint-report
-      - k8s-static-analysis
-      - k8s-unit-tests-with-coverage
-    uses: canonical/identity-credentials-workflows/.github/workflows/charm-build.yaml@v2.0.3
+    # Runs on merge, or on a PR labelled `test-arm` to verify arm64-specific fixes.
```
kayra1, line 89: I like the idea a lot, but i don't like that PR labels are shared with issue labels. who thought that made any sense at github? what do you think about maybe having the enabled arches in the commit description instead? this would be more flexible so you can pick the test on a commit by commit basis.

kayra1 (approved):
There could be improvements, but I like it. added small nits.

.github/workflows/main.yaml
saltiyazan: Thanks @kayra1 
I will merge this to get some changes in and test tomorrow with your idea

.github/workflows/grafana-dashboard-lint.yaml
saltiyazan: On the shared workflows?

PR #1015, opened 2026-07-02
fix: failure tolerance alert rule threshold

### LINKED ISSUE #1012: Vault incorrectly alerts for  VaultFailureToleranceRule
### Bug Description

On a fresh sunbeam deployment, 3 vault units are deployed. Which means we can only lose 1 unit to not lose the cluster. 

But the alerts are configured to trigger if `vault_autopilot_failure_tolerance <=1` which is always the case with sunbeam deploying a 3 units vault cluster.

This generate a false positive

### To Reproduce

sunbeam cluster deploy -m ... 
subbeam enable observability 

### Environment

Vault  1.18/stable       507
Sunbeam Noble Caracal 
COS 2/edge 
juju 2.6.23

### Relevant log output

```shell
vault_autopilot_failure_tolerance{instance="openstack_97c88ed0-2373-4d3f-839a-c993c4eeac89_vault_vault/0", job="juju_openstack_97c88ed0_vault_prometheus_scrape-0", juju_application="vault", juju_charm="vault-k8s", juju_model="openstack", juju_model_uuid="97c88ed0-2373-4d3f-839a-c993c4eeac89", juju_unit="vault/0"}
1
```

### Additional context

_No response_

chanchiwai-ray:
# Description

Update the failure alert rule threshold to `== 0` instead of the original `<=1`. This is because when the `vault_autopilot_failure_tolerance == 1`, the cluster is still healthy and can tolerate losing one healthy node, so this should not be considered as a critical alert (e.g. this is always the case for a 3 nodes vault cluster). 

On the other hand, if `vault_autopilot_failure_tolerance == 0`, this means the cluster cannot afford losing ANY node, and the operator should do something to prevent the bad case from happening (e.g. fix the unhealthy node or add more nodes). This should be considered as a critical alert.

# Checklist:

Closes: #1012


saltiyazan (changes requested):
Hi @chanchiwai-ray !
Thanks for the fix. However I would rather go with the solution you proposed in the issue description [here](https://github.com/canonical/vault-k8s-operator/issues/1012#issuecomment-4829702820). Because this can be considered a breaking change, anyone depending on these alerts will stop receiving them. So I would rather keep them coming but with `warning` severity and add an `==0` rule with `critical` severity.
Additionally, please note that the issue was opened against `1.18` so the fix must land in `release-1.18` too. But I can backport there once merged.

chanchiwai-ray:
Thanks @saltiyazan for the review, I've kept the two possible cases in `<=1` threshold; they now have slightly different alert names to distinguish the two cases. 

This looks something like this now for 3 nodes vault-k8s. 
image
@peppepetra does it looks right to you?

saltiyazan (approved):
Thanks for the work @chanchiwai-ray 
I will backport this to 1.18 too.

